PR #1856, opened 2014-08-12
Don't install insecurely without el-get-allow-insecure.

technomancy:
In most cases, we assume any connection is insecure unless the URL
starts with "https://", "$USERNAME@", or "ssh". There are a few
exceptions: I'm assuming all Emacswiki packages are insecure, and I
don't think we can know whether packages installed via Google Go are
secure or not.

### DIFF
--- a/el-get.info
+++ b/el-get.info
@@ -517,6 +517,12 @@ certain settings, particularly for portability.
      As above for ELPA, this directory stores a local list of emacswiki
      recipes. Defaults to `~/.emacs.d/el-get/el-get/recipes/emacswiki/'.
 
+`el-get-allow-insecure'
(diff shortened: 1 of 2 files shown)


technomancy:
It has come to my attention that just checking for HTTPS URLs is woefully inadequate for downloads which happen in-process: https://glyph.twistedmatrix.com/2015/11/editor-malware

I've disabled all in-emacs downloads, only allowing HTTPS git in my own fork, but maybe it should at least be documented that this setting does not actually prevent insecure downloads from happening?

npostavs:
See also #2287.

> but maybe it should at least be documented that this setting does not actually prevent insecure downloads from happening?

Sounds like a good idea.

Issue #1967, opened 2014-10-24
Merging bundle-el or similar functionality into el-get core?


priyadarshan:
I wanted to point out to `el-get` maintainers the `bundle` wrapper (https://github.com/tarao/bundle-el). I have been using it for some time and found it synergistic to `el-get`, simple to use and with a syntax in tune with `el-get` spirit.

Possibly within a future milestone, would it make sense at all to merge `bundle-el` (if the maintainer agrees), or incorporate a similar feature into `el-get` core?

npostavs:
Done in #1976.

PR #1976, opened 2014-11-01
Merging bundle-el into el-get 

tarao:
As we discussed in #1967, I would like to merge my [bundle-el](https://github.com/tarao/bundle-el) to el-get.

- `bundle.el` ~~and `eval-after-load-compile.el`~~ will be merged as `el-get-bundle.el` ~~and `el-get-eval-after-load-compile.el` respectively~~
- Related functions are added to `el-get.el` as `autoloads` (They are not `require`d since they are not needed by the el-get's core functionality)
- No documentation is provided: I'm not sure where to add
    - The simplest way is to copy [the original one](https://github.com/tarao/bundle-el/blob/el-get/README.md) to el-get's README?


npostavs:
> My only real concern before merging is that The el-get-bundle macro info section is a copy of the el-get-bundle docstring. This will lead to the same problem as the el-get-sources variable in which the manual and docstring copies have diverged, so I think we really shouldn't repeat that mistake.

Um sorry, I think this was actually already fixed by b4ee5553afde47e877ffd4dfd30d37a62347773c. I must have been looking at an old revision. Just speaking generally, the manual should have more details and example usage, while docstrings should have just the minimum summary of the behaviour.

> There can be many forms for a single package when you write multiple (el-get-bundle PACKAGE ...) and these may not be in the same file (when they are located in multiple files loaded from .emacs).

So multiple calls to `el-get-bundle` get merged? Knowing this, your code makes a lot more sense now. :) Perhaps you can add a note about that to the manual.

Okay, I think that's it for me. @tarao: feel free to (finally) install this. :)

tarao:
> the manual should have more details and example usage, while docstrings should have just the minimum summary of the behaviour

I get your point now.

> Perhaps you can add a note about that to the manual.

Added in 05605bd.

Now I merge this pull request. Thank you @npostavs, @dimitri for the review, and @priyadarshan, who have suggested merging `bundle-el` to `el-get`.

dimitri:
Awesome, thanks!

dimitri:
Should we include https://github.com/tarao/with-eval-after-load-feature-el/blob/master/with-eval-after-load-feature.el now?

priyadarshan:
Thank you so much!

A _very_ happy `el-get` user

PR #1999, opened 2014-11-20
Update commit hash for Emacs 23-compatible version of package.el

purcell:
I _believe_ this is the new hash for the same commit, though it's hard to be certain. In any case, the old URL was broken due to the rewriting of the Emacs git repo as part of the bzr->git migration.


npostavs:
Just leaving a note here, I've recently discovered that this version is not quite completely compatible with Emacs 23 because its `package-delete` uses the TRASH argument to `delete-directory`. Doesn't affect el-get (except that the compilation warning needs to be suppressed/ignored).

Issue #2005, opened 2014-11-23
Why doesn't el-get-reload use unload-feature?


npostavs:
`git blame` tells me @dimitri and @DarwinAwardWinner were the main authors of this; some discussion about it in #434. 

No mentions of `unload-feature`. I think it would make sense to use it...

npostavs:
> I doubt I have a wide enough package selection to test very effectively though.

A further problem is that even if a reload fails, I won't know whether it's *because* or *in spite of* calling `unload-feature`, and if it succeeds I won't know whether it's *because* or *regardless of*.

PR #2011, opened 2014-11-30
I think the &optional package-status-alist argument isn't needed

npostavs:
I noticed `el-get-merge-properties-into-status` had this `&optional package-status-alist` which was never actually used, except for the call in `el-get-do-init` which was only passing *its* `&optional package-status-alist`, and so on... There seems to be a whole bunch of functions with this optional argument that is never actually used anywhere. After removing them all, I get no compile errors/warnings about it, so I think this is safe.

```
* el-get-status.el (el-get-package-status-alist,
el-get-package-status-recipes, el-get-read-package-status,
el-get-read-package-status-recipe, el-get-with-status-sources,
el-get-read-cached-recipe, el-get-merge-properties-into-status):
* el-get.el (el-get-init, el-get-do-init, el-get-reload, el-get-remove,
el-get-checksum): Remove &optional package-status-alist argument.
```

### DIFF
--- a/el-get-status.el
+++ b/el-get-status.el
@@ -157,30 +157,28 @@
           else
           collect (cons p prop))))
 
-(defun el-get-package-status-alist (&optional package-status-alist)
+(defun el-get-package-status-alist ()
   "return an alist of (PACKAGE . STATUS)"
-  (loop for (p . prop) in (or package-status-alist
-                              (el-get-read-status-file))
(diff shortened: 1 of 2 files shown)


npostavs:
I'm holding off on installing this since it causes errors on update. I *think* it's only "online" update, that is, after restarting Emacs everything should be fine. Might be an interesting test case for #2005.

dimitri:
Thanks!

PR #2051, opened 2014-12-30
Add evil-org-mode recipe

f0086:


npostavs:
Thanks, squashed to trunk.

PR #2053, opened 2015-01-01
Return correct library from `:pkgname` for github based recipe

dochang:
If `:library` and `:feature` are not specified, `eval-after-load` will use `:pkgname` as its first argument.  For a github based recipe, the `:pkgname` has two parts: "username/reponame".  We need to remove the "username" part before passing `:pkgname` to `eval-after-load`.


npostavs:
Thanks, it's interesting that nobody noticed this before. Probably because many recipes had `:feature` which we've been removing if autoloads are provided.

PR #2062, opened 2015-01-06
add recipe for bts

aki2o:
Is bts-test.el compiled in this recipe condition?
If so, could you tell me how to exclude bts-test.el from compiled files?
bts-test.el is for test and requires other local packages for test.


dimitri:
See the docs for the `:compile` property of the recipe (full docs are available as Emacs Info).

npostavs:
squashed as 6096f4d250197ce169ed66c766a33b3fb7e192b8

PR #2064, opened 2015-01-09
Add recipe for with-eval-after-load-feature.

tarao:
This package not necessarily be a part of El-Get (since it is still useful to be used without El-Get) but at least a recipe should provided.

cf. #1976


dimitri:
Please merge it!

PR #2066, opened 2015-01-15
Create nim-mode

arunmib:
Adding nim-lang mode support.


npostavs:
Is there some reason you didn't make it `:type github`?

Also, it should work without `:features` since there are `;;;###autoloads`.

ruediger:
I made a PR for `nim-mode` as well: #2086 It should fix the issues

PR #2067, opened 2015-01-15
Add recipe for `arcadia-dot-el' package

sonelliot:
This package is for integration between Emacs and [Arcadia](https://github.com/arcadia-unity/arcadia-dot-el). Only just starting out, but it would be good to be able to install it with el-get!

> c:/cygwin64/home/Elliot/.emacs.d/el-get/el-get/recipes/arcadia.rcp: 0 error(s) found.

### DIFF
--- a/recipes/arcadia.rcp
+++ b/recipes/arcadia.rcp
@@ -1,0 +1,7 @@
+(:name arcadia
+ :description "Emacs integration with the Arcadia Unity project"
+ :type github
+ :features arcadia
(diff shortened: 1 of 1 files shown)


dimitri:
Can you test if that works without using the `:features` property?

npostavs:
I don't think `:username` works for `:type github` anymore, are you using the old stable branch (4.1)?.

npostavs:
closing for lack of response

PR #2069, opened 2015-01-17
add recipe for history.el

ngnmnfn:


npostavs:
I think it's better to name the recipe file history.rcp, like all the others.

PR #2071, opened 2015-01-18
Fix `el-get-bundle` callsite identifier generation

tarao:
The callsite identifier must include only valid characters for path names.  Characters other than `/` were replaced to `_` but this was not enough for example for GNU Emacs on Windows, where `:` appears after the drive letter of a callsite path.

### DIFF
--- a/el-get-bundle.el
+++ b/el-get-bundle.el
@@ -119,7 +119,7 @@
                (or (file-exists-p el-get-bundle-init-directory)
                    (make-directory el-get-bundle-init-directory t) t)))
     (let* ((path (file-name-sans-extension (expand-file-name load-file-name)))
-           (callsite (mapconcat #'identity (split-string path "/") "_"))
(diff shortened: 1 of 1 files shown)


npostavs:
`make-backup-file-name-1` deals with a similar problem, in addtion to replacing the drive name they also have
```elisp
	      ;; Replace any invalid file-name characters (for the
	      ;; case of backing up remote files).
	      (setq file (expand-file-name (convert-standard-filename file)))
```

tarao:
> `convert-standard-filename`

Done: f7b45e2

tarao:
I found that we have wrong description of this pull request.

> must include only valid characters for path names

"must include only valid characters for **file names**" is correct.

> Characters other than `/`

"Characters **no** other than `/`" is correct.

npostavs:
Feel free to rebase/reword/merge.

PR #2076, opened 2015-01-25
Add rake recipe

asok:
When I run `test/check-recipe.el recipes/rake.rcp` the output is blank.


npostavs:
> When I run test/check-recipe.el recipes/rake.rcp the output is blank.

Thanks, almost nobody ever tries running check-recipe.el beforehand (or at least they neglect to mention it).

PS: as of f540acc924db25e694419bc5a107a88cca2b909e it will say `0 warning/error(s) total.`

PR #2081, opened 2015-01-28
Added ocp-indent recipe.

sagotch:
ocp-indent tends to do a better job indenting ocaml code (e.g. exception in pattern matching).

Note: It requires you to install ocp-ident but that is not el-get's job.

### DIFF
--- a/recipes/ocp-indent.rcp
+++ b/recipes/ocp-indent.rcp
@@ -1,0 +1,5 @@
+(:name ocp-indent
+       :description "Automatic indentation with ocp-indent"
+       :type http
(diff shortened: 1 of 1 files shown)


dimitri:
Does it work without the `:features` property?

sagotch:
Even if I am not sure that I understand why, is does not work without `:features`.

By the way: I could not find any documentation for recipe writers. Could you spare any pointer about it?

dimitri:
Try M-x info el-get, there's a whole manual in there, browsable from within Emacs, with all you need to know.

PR #2082, opened 2015-01-30
* recipes/hydra.rcp: new recipe.

deftsp:


yyr:
Hmm, I was fast in merging this. It seems to depend on cl-lib. please `:depends` property.

deftsp:
done. https://github.com/dimitri/el-get/pull/2083

PR #2092, opened 2015-02-15
Update matlab-mode.rcp

lgiraldi:
The build command now works for different emacs versions


npostavs:
I removed the `shell-quote-argument` as it's not needed when using the `((<list of strings args>))` form.

PR #2105, opened 2015-03-01
Added opam-mode recipe.

sagotch:
Note: `features` are required, how to (and why) avoid it ?

### DIFF
--- a/recipes/opam-mode.rcp
+++ b/recipes/opam-mode.rcp
@@ -1,0 +1,4 @@
+(:name opam-mode
+       :description "Major mode for editing opam files"
+       :type github
+       :pkgname "sagotch/opam-mode.el")


npostavs:
It's preferred to autoload packages so that they can be installed without increasing startup load time. You can achieve this by adding `;;;###autoload` on the lines just before the main entry points to your package, in your case `(defun opam-mode ...` and `(defun opam-url-mode ...`.

You should also add `;;;###autoload` before `(add-to-list 'auto-mode-alist...)` calls so your mode will be activated (and loaded) when the user opens an applicable file.

sagotch:
Thanks! I'll try it and edit my recipe. **please do not merge**

sagotch:
`features` not required anymore. Thank you!

npostavs:
By the way, I don't think there's much point in `provide`ing features that aren't the same as the file name.

sagotch:
Why not?

`opam-mode` and `opam-url-mode` are meant to be used with different files, but to work together.

npostavs:
Because if you `(require 'opam-url-mode)` before `opam-mode.el` is loaded it won't work, it would search for `opam-url-mode.el`. Defining both modes is fine, but the require/provide thing really works with file names.

sagotch:
I thought that you needed to provide both mode to be able to use them... but you do not.

Noted, and fixed :+1:

PR #2106, opened 2015-03-02
add ac-php

xcwen:


npostavs:
I think the `:depends` need to be a single list (need to add this to check-recipe).

xcwen:
ok.

PR #2110, opened 2015-03-04
Fix el-get-git-compute-checksum

npostavs:
was failing with git-executable = `C:\Program Files (x86)\Git\cmd\git.exe`


dimitri:
Nice thanks! Is call-process recent in Emacs, should it warrant a compatibility note in the docs?

npostavs:
> Is call-process recent in Emacs

Nope, it's mentioned (i.e. already existed) in NEWS.18

PR #2121, opened 2015-03-07
Add dash to the list of dependencies of ebib

rlaboiss:
Please merge this commit into master.  It is strange that this has not been noticed before, since dash was added as a dependency of ebib over a year ago (joostkremers/ebib@4818d0c).  Thanks.


npostavs:
> It is strange that this has not been noticed before,

Probably because if you have dash installed already for another reason, there won't be any trouble.

PR #2122, opened 2015-03-09
Support checksums for http-tar and http-zip

cantora:
These changes were necessary for me to get http-tar and http-zip working with checksums.  

They are both configured to use the same checksum function as the http method, except they also run the verification process early (before the verification in the `el-get-post-install` function).  

I added the early checksum because it's risky to unzip/untar an archive that hasn't been verified. There might be a cleaner way to do this: it seems like putting the "generic" checksum verification in `el-get-post-install` is too late. It might be better to change the generic process such that there is a stage before "install" called "retrieve" and it models the abstract process of getting all the sources and verifying checksums. Then the zip and tar methods would only implement the unzip/untar process in the "install" stage (which comes after the download/checksum "retrieve" stage). Just an idea... let me know if I'm unclear.

### DIFF
--- a/el-get.el
+++ b/el-get.el
@@ -513,16 +513,11 @@ PACKAGE may be either a string or the corresponding symbol."
   (el-get-do-init package)
   (run-hook-with-args 'el-get-post-install-hooks package))
 
-(defun el-get-post-install (package)
-  "Post install PACKAGE. This will get run by a sentinel."
-  (let* ((sync             el-get-default-process-sync)
-         (type             (el-get-package-type package))
-         (hooks            (el-get-method type :install-hook))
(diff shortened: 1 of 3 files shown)


npostavs:
> There might be a cleaner way to do this: it seems like putting the "generic" checksum verification in el-get-post-install is too late. It might be better to change the generic process such that there is a stage before "install" called "retrieve" and it models the abstract process of getting all the sources and verifying checksums. Then the zip and tar methods would only implement the unzip/untar process in the "install" stage (which comes after the download/checksum "retrieve" stage).

Yeah, that is a good idea, `:type http` could use it too. Currently, if you download a bad .el file by http, you lose the good one.

PR #2124, opened 2015-03-13
Fix go get update issue

nanjj:


methods/el-get-go.el
```diff
@@ -22,30 +22,43 @@
...
                       :buffer-name ,name
                       :default-directory ,el-get-dir
                       :program ,el-get-go
-                      :args ("get" "-v" "-u" ,pkgname)
+                      :args ("get" ,pkgname)
                       :message ,ok
                       :error ,ko))
      post-install-fun)
-    (setenv "GOPATH" gopath)))
+    (make-symbolic-link (el-get-go-path gopath pkgname) pdir)))
```
npostavs: This won't work on Windows.

methods/el-get-go.el
```diff
@@ -22,30 +22,43 @@
...
+  (while
+      (and
+       (< i size)
+       (not
+        (file-directory-p
+         (expand-file-name
+          pkgname (expand-file-name "src" (elt v i))))))
+       (setq i (+ 1 i)))
+     (elt v i)))
+
```
npostavs: I think this function would clearer if written with `some` (aka `cl-some` though we can't use cl-lib).

npostavs:
@nanjj: ping?

methods/el-get-go.el
npostavs: I *think* this could be written:
```elisp
(defun el-get-go-path (gopath pkgname)
  "Find go path with PKGNAME."
  (locate-file (concat "src/" pkgname) (split-string gopath ":" t)
               nil #'file-directory-p))
```

npostavs:
@dougm: as the primary author of `el-get-go.el`, does this change make sense?

npostavs:
closing for lack of response

PR #2131, opened 2015-03-21
add some recipes

stardiviner:
I think it's correct now.


npostavs:
You can `push -f` to reuse the same pull request, by the way.

stardiviner:
That's really a good tip. Thanks

PR #2132, opened 2015-03-22
Add org-ioslide recipe.

akopytov:


npostavs:
Not sure why the travis test complains about whitespace in el-get.texi, when this pr doesn't modify el-get.txi at all...

akopytov:
I wondered the same thing. Thanks for merging!

npostavs:
Hmm, it might be taking the diff from the branch point (i.e. parent of 5845169223d86646dc0dad632c4dbb16be8adf0b which is 4e3b8b882d7a11230cb9a6587387713ccb4df438) to the merge point. It *ought* to diff from branch point to the "tip".

PR #2136, opened 2015-03-23
Add recipe for `org-ref'

d-p:


recipes/org-ref.rcp
```diff
@@ -0,0 +1,7 @@
+(:name org-ref
+       :website "https://github.com/jkitchin/org-ref"
+       :description "org-mode modules for citations, cross-references, bibliographies in org-mode and useful bibtex tools to go with it.
+"
```
npostavs: We don't need this newline in the `:description`.

d-p: Sorry, did not dobule checked.
By the way it passes the test: el-get-check-recipe.
I will resend a pull request soon.

PR #2141, opened 2015-04-01
magit moved there magit-svn into a different project

Narfinger:
It should be included in the recipies.


npostavs:
I think all of the fields from `:info` and after aren't needed for this package, can you try removing them?

tarsius:
You also need to add `:branch "master"`, see #2145.

PR #2144, opened 2015-04-02
Update magit.

Pitometsu:
please, merge to master.

### DIFF
--- a/recipes/magit.rcp
+++ b/recipes/magit.rcp
@@ -3,7 +3,7 @@
        :description "It's Magit! An Emacs mode for Git."
        :type github
        :pkgname "magit/magit"
-       :branch "next"
+       :branch "master"
        :depends (cl-lib dash)
        ;; use the Makefile to produce the info manual, el-get can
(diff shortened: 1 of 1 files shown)


Pitometsu:
According to https://github.com/dimitri/el-get/pull/2145
it should use :branch master as well yet.

Pitometsu:
@npostavs please, merge `:build/darwin` as well!

npostavs:
oh, oops, didn't notice you had added that. I just moved the `EMACSBIN` param into the other `:build/`s instead, it shouldn't cause any harm. 450c6406cf7567dc01eb41314178d14d87b54ce7

Pitometsu:
```
make: *** No rule to make target `docs'.  Stop.
```
Please, add `"lisp"` instead of `"docs"` too.

npostavs:
huh? Are you installing `next`? The recipe as it is right now works for me.

Pitometsu:
@npostavs oh, sorry, u right, it's ok now.

PR #2145, opened 2015-04-02
build magit from the master branch

tarsius:
Yesterday I have just released 1.4.0 and am now preparing to release 2.1.0.
The new HEAD branch is next, but I would like the El-get packages to be
build from the master branches for a little longer. I will switch to the next
branches approximately to weeks before 2.1.0 is actually released.


tarsius:
Also see https://github.com/milkypostman/melpa/pull/2640.

PR #2147, opened 2015-04-04
The deferred package should be installed along with request

hadenpike:


npostavs:
You're deleting ` recipes/simplenote.rcp`, is that on purpose?

hadenpike:
No. I thought I had reverted that change. In any case, it should not be removed.

PR #2148, opened 2015-04-04
The deferred packageshould be installed along with Request

hadenpike:
This is the correct version of the earlier pull request that does not remove any files.


npostavs:
Thanks.

P.S. you can `git push -f` without closing the pull request to reuse it.

PR #2149, opened 2015-04-05
New recipe: simplenote2

hadenpike:
I don't know the policy for deprecated recipes,
so I didn't remove simplenote.el.

### DIFF
--- a/recipes/simplenote2.rcp
+++ b/recipes/simplenote2.rcp
@@ -4,4 +4,5 @@
        :depends (request)
        :type github
        :pkgname "alpha22jp/simplenote2.el"
-       :post-init (simplenote2-setup))
+       :post-init (simplenote2-setup)
+       :lazy t)
--- a/recipes/simplenote2.rcp
+++ b/recipes/simplenote2.rcp
@@ -4,7 +4,4 @@
        :depends (request)
        :type github
        :pkgname "alpha22jp/simplenote2.el"
-       :features simplenote2
-       :load    "simplenote2.el"
-       :compile "simplenote2.el"
        :post-init (simplenote2-setup))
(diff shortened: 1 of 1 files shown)


recipes/simplenote2.rcp
```diff
@@ -0,0 +1,10 @@
+(:name simplenote2
+       :website "https://github.com/alpha22jp/simplenote2.el#readme"
+       :description "A new version of emacs helper to interact with Simplenote"
+       :depends (request)
+       :type github
+       :pkgname "alpha22jp/simplenote2.el"
+       :features simplenote2
```
dholm: The package has `autoload` cookies, you don't need `features`.

recipes/simplenote2.rcp
```diff
@@ -0,0 +1,10 @@
...
+       :load    "simplenote2.el"
```
dholm: You don't need to load it, it comes with `autoload` cookies.

recipes/simplenote2.rcp
```diff
@@ -0,0 +1,10 @@
...
+       :compile "simplenote2.el"
```
dholm: `:compile` is only needed for special cases.  I don't think you need it for this package as it looks pretty straight-forward.

recipes/simplenote2.rcp
```diff
@@ -0,0 +1,10 @@
...
+       :post-init (simplenote2-setup))
```
dholm: Perhaps you can make `simplenote2-setup` run after `simplenote2` has been loaded?  Otherwise you will force the package to be loaded every time Emacs is started.

hadenpike:
Done.

npostavs:
> Perhaps you can make simplenote2-setup run after simplenote2 has been loaded?

This could be done by adding `:lazy t` to the recipe.

hadenpike:
Done.

PR #2151, opened 2015-04-07
just warn about install failures

### LINKED ISSUE #1625: el-get to fail gracefully
Marmalade has been up and down this last week, and el-get-init-and-install has not been failing gracefully. Either it times out with a bad gateway and my init.el stops, or it cannot get the package from the server, and stops.

I would like e-get to fail gracefully (with warnings etc) when this kind of thing happens.

npostavs:
I was at first a bit unsure the possiblity of cascading errors would be a problem, but after seeing this in action in `use-package` I think it's much better to chance a few extra error messages in the `*Warnings*` buffer than failing to get through the init.el.

fixes #1625, though in my tests network failures seem to result in very long timeouts, so I doubt I'd have the patience to wait for a warning message.

### DIFF
--- a/el-get-core.el
+++ b/el-get-core.el
@@ -41,6 +41,17 @@ the original object."
 (defun el-get-verbose-message (format &rest arguments)
   (when el-get-verbose (apply 'message format arguments)))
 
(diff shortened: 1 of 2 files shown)


dimitri:
Thanks!

PR #2175, opened 2015-05-23
recipe: powershell.el

jschaf:
I combined powershell.el and powershell-mode.el that were floating around emacs-wiki. I cleaned up the documentation and reorganized some of the code.

I received permission from the creator to re-use the name powershell to create a canonical package.

https://github.com/jschaf/powershell.el

A little bit of history is available in my Melpa pull request: https://github.com/milkypostman/melpa/pull/1308


yyr:
Thanks

PR #2192, opened 2015-06-16
Fix dependency of elscreen

masutaka:
Nowadays elsceen does not require apel.
https://github.com/emacs-jp/elscreen/blob/master/elscreen.el#L30-L34

By the way, elscreen required apel at 2639154.


masutaka:
Thanks!

PR #2196, opened 2015-06-20
Fix missing let-alist.el

masutaka:
See http://git.savannah.gnu.org/cgit/emacs.git/commit/lisp?id=02617f15075177cd5a390dd13bd3fc6b49e08fad


masutaka:
thx!

PR #2203, opened 2015-07-02
Remove all shell-interpolation from build commands (v2)

### LINKED ISSUE #1959: Dictionary recipe gives warning on install
*Warning (emacs): Build command "make EMACS=/usr/local/bin/emacs" in package "dictionary" will be shell-interpolated. To bypass shell interpolation, the recipe for "dictionary" should specify build commands as lists of strings instead.*

Not sure what that means, but it occurs on a fresh install.  It appears to install the plugin, but this warning appears on first time install.

npostavs:
The travis tests will now check for shell interpolation in build recipes, so that should prevent any new ones from creeping in.

fix #1959.

This time from right the repository, and without crlfs.


dimitri:
Thanks for your work on that, it's appreciated!

npostavs:
I made the checks more thorough and caught some more interpolation. There are 2 recipes not checked due to their using `unsafep` lisp functions (listed by passing `-W:debug`):

```console
$ emacs -Q -L . -batch -l el-get-check -f el-get-check-recipe-batch -W:debug -Wno-features -Wno-github -Wno-autoloads recipes/ 
recipes/helm.rcp:Debug: :build/windows-nt is unsafep: (function update-directory-autoloads)
recipes/emacs-goodies-el.rcp:Debug: :build is unsafep: (function shell-command-to-string)
2 warning/error(s) total.
```

`helm.rcp :build/windows-nt` does actually generate autoloads via lisp, so we don't want to evaluate it during a check. `emacs-goodies-el.rcp :build` uses `(shell-command-to-string "find . -name '*.make'")` which is probably safe(ish), but I don't think it makes sense to analyse shell commands for safety.

PR #2204, opened 2015-07-02
with magit 2.1.0 the info file moved to Documentation/

tosmi:


npostavs:
whoops, missed that. Thanks.

PR #2205, opened 2015-07-04
Add org-opml recipe.

deftsp:


npostavs:
I would suggest using
```
       :prepare (progn
                  (autoload 'set-buffer-file-format-to-opml "org-opml")
                  (add-hook 'org-mode-hook 'set-buffer-file-format-to-opml)
                  (add-to-list 'auto-mode-alist '("\\.opml$" . org-mode))
                  (el-get-envpath-prepend "PATH" default-directory)
                  (add-to-list 'format-alist '(opml "Outline Processor Markup Language"
                                                    "<[?]xml version=\"1.0\"[^>]*[?]>[\n]?.*[\n]?.*[\n]?<opml version=\"[1|2].0\">"
                                                    "opml2org.py" opml-encode t)))
       :post-init (remove-hook 'find-file-hooks 'set-buffer-file-format-to-opml)
```

Upstream's use of `find-file-hooks` seems like overkill when they could use `org-mode-hook` instead,
and I don't think  hardcoding `"~/src/org-opml/opml2org.py"` can work with an el-get installed package at all, so it's important to add to `format-alist` before loading the package.

deftsp:
Sorry, I just use org-opml to convert org file to opml, have not edited opml file directly.

PR #2208, opened 2015-07-06
add let-alist as dependency to pdf-tools recipe

### LINKED ISSUE #2207: pdf-tools recipe does not work
It complains about no let-alist. Installing manually (`package-install-file`) as instructed at https://github.com/politza/pdf-tools works. Doing:

    (el-get-bundle let-alist)
    (el-get-bundle pdf-tools)

however, works too. I'm not sure if the recipe should have `:depends (let-alist)`, and I'm not sure how to check for that. [Link to the recipe](https://github.com/dimitri/el-get/blob/master/recipes/pdf-tools.rcp).

As I understood it, let-alist was already a built in package, so maybe this is just me. Does the recipe work for you?

vyp:
Closes #2207.

---

(I hope?)

### DIFF
--- a/recipes/pdf-tools.rcp
+++ b/recipes/pdf-tools.rcp
@@ -1,9 +1,11 @@
 (:name pdf-tools
        :description "pdf tools"
        :type github
+       :minimum-emacs-version "24.3"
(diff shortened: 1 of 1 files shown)


npostavs:
Looks good, thanks.

Issue #1471, opened 2013-12-12
dependencies within main .el file not respected


aspiers:
Maybe I'm missing something, but it looks to me like `el-get` does not respect the contents of the `Package-Requires` header within the package's main `.el` file.  For example, I just did `M-x el-get-install pkg-info` and it failed to load `epl` despite `pkg-info.el` containing the following: 

```
;;; pkg-info.el --- Information about packages       -*- lexical-binding: t; -*-

;; Copyright (C) 2013  Sebastian Wiesner

;; Author: Sebastian Wiesner <lunaryorn@gmail.com>
;; URL: https://github.com/lunaryorn/pkg-info.el
;; Keywords: convenience
;; Version: 0.5-cvs
;; Package-Requires: ((dash "1.6.0") (epl "0.4"))
```

If I do `M-x el-get-find-recipe-file pkg-info` then I get

```
(:name pkg-info
       :description "Provide information about Emacs packages."
       :type github
       :pkgname "lunaryorn/pkg-info.el"
       :depends s)
```

and `M-x el-get-describe pkg-info` gives:

```
pkg-info is an `el-get' package.  It is currently installed [update][remove].

Website: https://github.com/lunaryorn/pkg-info.el.git
Description: Provide information about Emacs packages.
Dependency: `s'.
The default installation method is github.
Installed in `/home/adam/.emacs.d/el-get/pkg-info/'.

Full definition in `/home/adam/.emacs.d/el-get/el-get/recipes/pkg-info.rcp':
(:name pkg-info :description "Provide information about Emacs packages." :type github :pkgname "lunaryorn/pkg-info.el" :depends s)

[back]
```

so clearly `el-get` thinks it depends on `s` when in actual fact it depends on `dash` and `epl`.

Again, I may be misunderstanding something here, but if I'm right, this is a fundamentally broken approach, because upstream package dependencies will naturally change on a regular basis, so any model which violates DRY and expects the dependencies to be updated in more than one place will inevitably lead to frequent breakage.  Sure, there may be occasions where it is necessary to model dependencies in the recipe in _addition_ to those in the package header (e.g. if upstream fails to correctly model the dependencies), but this is not a good reason to fail to honour the package header.

npostavs:
[...]

PR #2212, opened 2015-07-10
Add auto update dependency functions

npostavs:
see #1471. Not a full solution, but maybe it'll help a bit and a full solution will use parts of this.

### DIFF
--- a/el-get-core.el
+++ b/el-get-core.el
@@ -216,6 +216,15 @@ entry."
   (if (string-match-p el-get-no-shell-quote arg) arg
     (shell-quote-argument arg)))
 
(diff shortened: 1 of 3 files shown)


dimitri:
Looks nice!

Should we make it interactive (as in a minor mode available for recipe files, or a major mode even, I don't know), or just automatic someway?

npostavs:
Well it's basically 1 command to run, it doesn't seem worth making a whole mode. Do we have an existing mode for recipes? If so we could add a binding.

dimitri:
we don't, I was trying to trick you into getting that started ;-)

PR #2213, opened 2015-07-13
add recipe for rust racer

LefterisJP:


recipes/rust-racer.rcp
```diff
@@ -0,0 +1,15 @@
...
+       :pkgname "phildawes/racer"
+       :description "Rust code completion and code navigation"
+       :build '(("cargo" "build" "--release"))
+       :depends (rust-mode company-mode)
+       :load-path "editors/emacs"
+       :post-init (progn
+                    (setq racer-cmd (concat default-directory "/target/release/racer"))
+                    (eval-after-load "rust-mode" '(require 'racer))
+                    (add-hook 'rust-mode-hook
+                              '(lambda () (racer-activate)))))
```
npostavs: You should replace the `'(lambda...)` with just `#'racer-activate`. And I think the `(eval-after-load "rust-mode" '(require 'racer))` shouldn't be needed since `racer-activate` has an `;;;###autoload` cookie.

LefterisJP: yeah I think you are right. Let me amend the commit.

npostavs:
Thanks, picked to trunk. (I moved the `add-hook` to `:prepare` so it will work in lazy mode too)

PR #2217, opened 2015-07-17
update jade-mode recipe to use github instead of elpa

mk0x9:
Pick jade-mode from upstream, ELPA package is outdated.


npostavs:
Thanks, I changed to `:type github`.

PR #2218, opened 2015-07-18
Update sx.rcp

vermiculus:
See https://gitter.im/vermiculus/sx.el?at=55a9a03810521a7e52558ac8

PR #2219, opened 2015-07-18
Fix sx recipe

mk0x9:
sx.el [requires let-alist](https://github.com/vermiculus/sx.el/blob/master/sx.el#L9) and requiring sx-tab is mandatory, because el-get placing all autoloads not in package's directory, but in one in `~/.emacs.d/el-get`.

### DIFF
--- a/recipes/sx.rcp
+++ b/recipes/sx.rcp
@@ -2,5 +2,5 @@
        :description "Stack Exchange for Emacs."
        :website "https://github.com/vermiculus/sx.el"
        :type github
-       :depends (markdown-mode)
+       :depends (markdown-mode let-alist)
        :pkgname "vermiculus/sx.el")


vermiculus:
This should probably be preferred over #2218, then :smile:

mk0x9:
Oh, didn't saw yours. :-)

npostavs:
Ah, saw this one afterwards, but as I said in gitter:

> About the autoload thing, shouldn't the .el files be on the load path already so just `(autoload 'sx-tab-hot "sx-tab" nil t)` should work?

mk0x9:
Hey, [autoloads have been fixed](https://github.com/vermiculus/sx.el/pull/299) according to your advice. So no `:features` anymore.

vermiculus:
Yep, vermiculus/sx.el#299

npostavs:
Great, thanks.

PR #2222, opened 2015-07-21
Magit recipe updates [Please Test]

npostavs:
I'd appreciate if someone with access to a linux and/or bsd box can test this out, currently I can only test on Windows.

---

@dholm: Could you explain exactly what problem 75edc70c4dc9ad0a821b8a00a5df1c5e10413e3c is supposed to fix? I don't believe it actually has any effect because there is no where in the code that checks for `:provide`.

### DIFF
--- a/el-get-build.el
+++ b/el-get-build.el
@@ -49,7 +49,8 @@ Otherwise, use `:build/SYSTEM-TYPE' or `:build'."
               ;; expression that evaluates to a command list, rather than a
               ;; literal command list.
               (if (symbolp (car raw-build-commands))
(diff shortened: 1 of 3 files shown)


dholm:
@npostavs The issue is that MELPA has a package for `with-editor` that only provides that particular component.  If you install a package that depends on `with-editor` the MELPA version would get pulled in and cause issues for other packages depending on whether `with-editor` or `magit` is first in your `load-path`.

npostavs:
@dholm: okay, but adding `:provide` does nothing to solve that, right?

npostavs:
I've merged this without the last commit as I've realized we will need `:provide` to properly fix #1471.

PR #2230, opened 2015-08-08
[new] celery

ardumont:
Hello,

Checks done below.

```command
M-x el-get-check-recipe RET
/home/tony/repo/perso/el-get/recipes/celery.rcp: 0 error(s) found.
```

Also installed from recipe and ok.

```txt
el-get install celery
el-get: el-get-build celery: byte-compile ok.
el-get: updating autoloads for celery
Generating autoloads for celery/celery.el...done
```

Cheers,

### DIFF
--- a/recipes/celery.rcp
+++ b/recipes/celery.rcp
@@ -1,0 +1,5 @@
+(:name celery
+       :description "a minor mode to draw stats from celery and more?"
+       :type github
+       :pkgname "ardumont/emacs-celery"
+       :depends (dash s deferred))


recipes/celery.rcp
```diff
@@ -0,0 +1,5 @@
+(:name celery
+       :description "a minor mode to draw stats from celery and more?"
+       :type github
+       :pkgname "ardumont/emacs-celery"
+       :depends (dash-functional s deferred))
```
npostavs: el-get only has a `dash` package, not `dash-functional` (the dash package installs the whole repository including the `dash-functional` file). I'm a bit surprised it installed successfully.

ardumont: > el-get only has a dash package, not dash-functional (the dash package installs the whole repository including the dash-functional file). I'm a bit surprised it installed successfully.

Well, i don't know. I have the dependencies installed.

Maybe it's because it's an melpa el-get?



I'll modify accordingly then.

Thanks for the feedback.

npostavs: > Maybe it's because it's an melpa el-get?

Oh right, that's it. Forgot about that.

PR #2234, opened 2015-08-22
Add Perforce p4.el recipe

crzysdrs:
This adds p4.el from gareth-rees/p4.el which enables common Perforce commands in emacs.


dimitri:
Does it work if you remove the `:load` and `:autoload` properties of the recipe?

crzysdrs:
Removing the :autoloads works just fine but without the :load the added commands won't be available without additional elisp in the user's initialization file.

dimitri:
Thanks for testing and explanations!

PR #2237, opened 2015-08-25
add type of source in list of packages

seblemaguer:
I propose to add the type of the source in the package list. I used this information to solve this problem:  there were some conflicts between a package proposed by "elpa" which depends on another one proposed by github. The version were not matching which leads to an inconsistency.


npostavs:
Thanks, merged after narrowing the "Type" column a bit more.

Issue #2232, opened 2015-08-15
Installing some packages with el-get-bundle will fail at newxt launch


priyadarshan:
When I install a package without regular recipe, like

```
(el-get-bundle doitian/ido-complete-space-or-hyphen)
```
`el-get` installs it fine, and I am able to use it.

Adding the above line to `init.`le, next time Emacs is launched, I get
```
error: El-get can not find a recipe for package "ido-complete-space-or-hyphen"
```

`.status.el` contains the following lines

```
(ido-complete-space-or-hyphen status "installed" recipe
			       (:name ido-complete-space-or-hyphen :type github :pkgname "doitian/ido-complete-space-or-hyphen" :after nil))
```
;;;***

;;;### (autoloads nil "ido-complete-space-or-hyphen/ido-complete-space-or-hyphen"
;;;;;;  "ido-complete-space-or-hyphen/ido-complete-space-or-hyphen.el"
;;;;;;  (21967 21428 0 0))
;;; Generated autoloads from ido-complete-space-or-hyphen/ido-complete-space-or-hyphen.el

(autoload 'ido-complete-space-or-hyphen-enable "ido-complete-space-or-hyphen/ido-complete-space-or-hyphen" "\
Enable ido-complete-space-or-hyphen

\(fn)" t nil)

(autoload 'ido-complete-space-or-hyphen-disable "ido-complete-space-or-hyphen/ido-complete-space-or-hyphen" "\
Disable ido-complete-space-or-hyphen

\(fn)" t nil)
```

**Comment by priyadarshan** [comment]
Adding some details:

Emacs version:

`GNU Emacs 24.5.1 (x86_64-apple-darwin13.4.0, NS apple-appkit-1265.21) of 2015-04-10 on builder10-9.porkrind.org`

Minumum content of `init.el` showing that behaviour:

```
(add-to-list 'load-path "~/.emacs.d/el-get/el-get")

(unless (require 'el-get nil 'noerror)
  (with-current-buffer
      (url-retrieve-synchronously
       "https://raw.githubusercontent.com/dimitri/el-get/master/el-get-install.el")
    (goto-char (point-max))
    (eval-print-last-sexp)))

(add-to-list 'el-get-recipe-path "~/.emacs.d/el-get-user/recipes")
(el-get 'sync)

(el-get-bundle doitian/ido-complete-space-or-hyphen)
```

**Comment by priyadarshan** [comment]
I think I have found the reason:

If the line
```
(el-get 'sync)
```

[...]

PR #2238, opened 2015-09-02
Update safe properties in .status.el (whitelist approach)

### LINKED ISSUE #1817: el-get-update doesn't update .status.el
Causes "obsolete integer for :builtin" warning even though package.rcp has been updated.
Also a problem after removing `git-modes` from magit `:depends`, throws error about not finding git-modes.

----

Redundancy? `el-get-update-requires-reinstall`

```elisp
(defun el-get-do-init (package &optional package-status-alist)
...
  (when el-get-auto-update-cached-recipes
    (el-get-merge-properties-into-status package package-status-alist :noerror t))
...
      (let* ((el-get-sources (el-get-package-status-recipes))
```

reinit safe
```
:load-path
:info
:load
:features
:library
:prepare
:before
:post-init
:after
:lazy
:website
:description
```

update safe
```
:depends
:build
:build/system-type
:compile
:checksum
:checkout
:options
:localname
```

requires reinstall
```
:pkgname
:type
:builtin ; safe if not crossing Emacs version
:branch
:url ; :url is update safe for http
:module
:repo
:shallow
:submodule
;; actually, :localname is update safe
```

npostavs:
Fixes #1817.

This extends the whitelist with a 2nd update whitelist, with special cases for `:builtin` (safe if we haven't "crossed" the Emac version) and `:url` (safe for some recipe types). 

I'm actually considering switching to a blacklist approach instead. Consider that removing (or adding) properties like `:provide` that are unknown to el-get is obviously perfectly safe. Essentially, properties that are not used by a type's update method are safe to change.

The downside of blacklists is it's more dangerous if we get the list wrong of course. But, leaving a stale recipe is also bad if we want to use it (eg a solution for #2232 might want to do that).

Thoughts?

### DIFF
--- a/el-get-core.el
+++ b/el-get-core.el
@@ -140,11 +140,13 @@ already-defined method DERIVED-FROM-NAME."
 ;; interface to write such structures as raw elisp.
 ;;
 ;;;  "Fuzzy" data structure conversion utilities
-(defun el-get-as-string (symbol-or-string)
-  "If STRING-OR-SYMBOL is already a string, return it.  Otherwise
-convert it to a string and return that."
-  (if (stringp symbol-or-string) symbol-or-string
-    (symbol-name symbol-or-string)))
+(defun el-get-as-string (obj)
+  "Return OBJ as a string."
+  (cond
+   ((stringp obj) obj)
+   ((symbolp obj) (symbol-name obj))
+   ((numberp obj) (number-to-string obj))
+   (t (error "Can't convert %S to string." obj))))
 
 (defun el-get-as-symbol (string-or-symbol)
   "If STRING-OR-SYMBOL is already a symbol, return it.  Otherwise
--- a/el-get-status.el
+++ b/el-get-status.el
@@ -260,53 +260,50 @@ that recipe in the el-get status file.")
     ,@el-get-status-init-whitelist)
   "Properties than can be updated by `el-get-update'.")
 
-(defun el-get-classify-new-properties (operation source newprops)
-  "Classify NEWPROPS into two groups: need update/disallowed.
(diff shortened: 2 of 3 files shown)


dimitri:
Not sure about the backlist idea here... the current tradeoff is towards making it as easy as possible to author recipes, at the expenses of loose control and sometimes harder to debug situations... that said, I'm not doing enough debugging here to have a horse in this race, so I'll trust you to make the right choice!

npostavs:
> the current tradeoff is towards making it as easy as possible to author recipes, at the expenses of loose control and sometimes harder to debug situations

Changing to a blacklist shouldn't affect recipe authoring at all. The difficulty would be in writing the blacklist itself, which has to be different per el-get `:type`. I think ultimately blacklisting is correct because changing unknown properties *is* safe, but since it's going to need more work I will go with this whitelist approach for now.

Or maybe it's better to just add a manual override to `el-get-merge-properties-into-status`, I've just noticed 
```
Warning (el-get): Must reinstall `undo-tree' to modify its cached recipe
  adding:   (:url "http://www.dr-qubit.org/git/undo-tree.git/")
  removing: (:url "http://www.dr-qubit.org/git/undo-tree.git")
```
To realize that is safe would require some url parsing, which really seems like overkill...

npostavs:
Another thing to consider is updating only some of the properties, currently it's all-or-nothing so some property changes deemed "unsafe" can hold the rest "hostage". I just hit this now after updating magit, the `:load-path` wasn't updated (because of some other property changes) causing a `Cannot open load file: no such file or directory, with-editor` error. Only current way to fix is `el-get-reinstall`.

PR #2254, opened 2015-10-02
Added vimish-fold.

nloyola:
This package mostly mimics text folding feature found in Vim.


npostavs:
I just added depend info according to [vimish-fold.el][1] (I skipped `cl-lib` because that seems redundant given the requirement for Emacs 24.4).

[1]: https://github.com/mrkkrp/vimish-fold/blob/9d07ce65aacd6a2baa6f219cb41e2ca0e0ab6bee/vimish-fold.el#L10

nloyola:
Thanks.

PR #2269, opened 2015-10-21
added recipe for jenkins mode

gizmomogwai:
see https://github.com/rmuslimov/jenkins.el


npostavs:
Is there a particular reason for `:shallow nil`? Also, suggest you use `:type github`.

npostavs:
I cherry-picked to trunk with a small modification (`:pkgname` instead of `:url`): 353ebabc8c78d594a00f693305b35a2f1e885265

gizmomogwai:
thanks a lot!

PR #2272, opened 2015-10-25
Update CIDER's deps

bbatsov:
We've recently dropped `dash.el` in favour of `seq.el`.


iced:
Recipe for seq:

(:name seq
       :description "Sequence manipulation library for Emacs."
       :type github
       :pkgname "NicolasPetton/seq.el")

PR #2274, opened 2015-10-27
Add recipe for seq

vincentbernat:
Upstream says "seq.el", but MELPA says "seq". This is builtin in Emacs
25.


npostavs:
Thanks, this is needed for #2272 as well.

PR #2277, opened 2015-10-28
Add init-open-recentf recipe

zonuexe:
Add recipe for [zonuexe/init-open-recentf.el](https://github.com/zonuexe/init-open-recentf.el)


zonuexe:
Thank you!

PR #2281, opened 2015-11-03
Rust-racer has been removed, using emacs-racer instead

Tsgzj:


npostavs:
Is `emacs-racer.rcp` supposed to be instead of `rust-racer.rcp` or in addition to?

Also, please squash this into a single commit, as there seems to be a bunch of changes that all just cancel each other out.

Tsgzj:
`rust-racer.rcp` used to work but no longer exists since they changed [racer](https://github.com/phildawes/racer).

PR #2283, opened 2015-11-04
Fix for users who disable GnuTLS

### LINKED ISSUE #2282: Emacs 24.3 (9.0) OSX fails on gnutils-available-p
```
(unless (require 'el-get nil 'noerror)
  (with-current-buffer
      (url-retrieve-synchronously
       "https://raw.github.com/dimitri/el-get/master/el-get-install.el")
    (goto-char (point-max))
    (eval-print-last-sexp)))
```

Evaluating the above in my .emacs causes the following to be thrown:
```
Symbol's function definition is void: gnutls-available-p
```

changing `"https://raw.github.com/dimitri/el-get/master/el-get-install.el"` to `"https://raw.github.com/dimitri/el-get/4.stable/el-get-install.el"` fixes it for me.

syohex:
Emacs which is built with '--without-gnutls' does not have
gnutls-available-p.

This ~~is related to~~ fixes #2282.
CC: @hershaw

### DIFF
--- a/el-get-install.el
+++ b/el-get-install.el
@@ -21,7 +21,7 @@
         (or (bound-and-true-p el-get-dir)
             (concat (file-name-as-directory user-emacs-directory) "el-get")))))
 
-  (unless (gnutls-available-p)
+  (unless (and (fboundp 'gnutls-available-p) (gnutls-available-p))
     (display-warning
      'el-get
      (concat "Your Emacs doesn't support HTTPS (TLS)"


hershaw:
Well that was fast. Thanks @syohex!

npostavs:
> Emacs which is built with '--without-gnutls' does not have
> gnutls-available-p.

Thanks, I was not aware of that.

PR #2289, opened 2015-11-14
Ert tests

### LINKED ISSUE #739: Change test file naming scheme?
We currently name tests based on issue numbers. This makes it hard to find a test for a specific feature that you know exists. I propose to rename the tests with short descriptions of what is being tested, and ensure that the test file links to any associated issues on github. Any objections? I can leave the issue number in the file name as well.

### LINKED ISSUE #1691: Tests. Like, yes, we should run them!
#### Background

The travis CI test script, as of e868142, only reports the result of byte compilation. Before it attempted to run the scripts under `tests`, but failed because `ert` isn't installed in emacs 23.

#### TODO

- [ ] https://github.com/rolandwalker/emacs-travis has useful templates for emacs travis tests across several versions. https://github.com/rejeep/evm may be useful too. Incorporate one or both of these.
- [ ] It's better to have tests that don't rely on network access, for both reliability and speed. Possibly related: #969. 
- [ ] As mentioned in #739, the test names could be nicer (fix this along with previous point).
- [x] It would be good to somehow test the recipes themselves, at least applying `el-get-check-recipe`.

npostavs:
Okay, I'm finding #2264 really needs some tests with it, so it's time to actually get a bit more serious about the ert tests.

Extends/replaces/ closes #969. We'll need to take another look at the failure, but for now I marked it as `:expected-failure`.

Close #739, it's basically too boring to go over all of the existing issue replicators so I can't see it ever getting done; I'm just moving them to a subdirectory.

Sort of fixes #1691.


dimitri:
Thanks!

PR #2300, opened 2015-11-26
Add recipe for prop-menu, update idris-mode deps

mcclurmc:
Idris mode reqires prop-menu to be installed, but that package wasn't availablein el-get. This commit adds a recipe for prop-menu, and updates the idris-mode recipe to depend on prop-menu.

### DIFF
--- a/recipes/prop-menu.rcp
+++ b/recipes/prop-menu.rcp
@@ -1,0 +1,6 @@
+(:name prop-menu
+      :website "https://github.com/david-christiansen/prop-menu-el"
+      :description "Compute pop-up menus from text and overlay properties "
+      :type github
+      :pgkname "david-christiansen/prop-menu-el"
+      :features prop-menu)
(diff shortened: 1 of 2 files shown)


recipes/prop-menu.rcp
```diff
@@ -0,0 +1,6 @@
+(:name prop-menu
+      :website "https://github.com/david-christiansen/prop-menu-el"
+      :description "Compute pop-up menus from text and overlay properties "
+      :type github
+      :pgkname "david-christiansen/prop-menu-el"
+      :features prop-menu)
```
npostavs: I think `:features` should not be needed becaus this is a library type package, i.e. other packages will `require` it as needed.

mcclurmc: Thanks @npostavs, I've removed `:features` as you have suggested.

npostavs:
Thanks, squashed to `master`.

PR #2304, opened 2015-12-03
Update let-alist recipe

### LINKED ISSUE #2302: Recipe for let-alist fails with 404
(:name let-alist
       :description "Easily let-bind values of an assoc-list by their names."
       :builtin "25.0.50"
       :type http
       :url "http://git.savannah.gnu.org/cgit/emacs/elpa.git/plain/packages/let-alist/let-alist.el")

Contacting host: git.savannah.gnu.org:80
Error (el-get): while installing let-alist: could not fetch URL http://git.savannah.gnu.org/cgit/emacs/elpa.git/plain/packages/let-alist/let-alist.el: error http (404)

syohex:
let-alist was removed from elpa git repository because it is core package now.

This fixes #2302.

### DIFF
--- a/recipes/let-alist.rcp
+++ b/recipes/let-alist.rcp
@@ -1,5 +1,5 @@
 (:name let-alist
        :description "Easily let-bind values of an assoc-list by their names."
        :builtin "25.0.50"
-       :type http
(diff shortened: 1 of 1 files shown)


npostavs:
Thanks.

PR #2317, opened 2016-01-10
Add recipe for duckpan.el

GuiltyDolphin:
Basic GitHub recipe.

Output from `check-recipe.el`:

screenshot from 2016-01-10 20 57 28

(I don't have any autoload cookies).

### DIFF
--- a/recipes/duckpan.rcp
+++ b/recipes/duckpan.rcp
@@ -1,8 +1,7 @@
 (:name duckpan
        :description "Tools for working with DuckDuckHack development"
        :type github
-       :pkgname "duckpan-emacs"
-       :username "GuiltyDolphin"
+       :pkgname "GuiltyDolphin/duckpan-emacs"
        :branch "master"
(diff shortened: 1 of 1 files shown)


recipes/duckpan.rcp
```diff
@@ -0,0 +1,8 @@
+(:name duckpan
+       :description "Tools for working with DuckDuckHack development"
+       :type github
+       :pkgname "duckpan-emacs"
+       :username "GuiltyDolphin"
```
npostavs: `:username` is not supported since a couple of years I think, you should use `:pkgname "GuiltyDolphin/duckpan-emacs"` instead

npostavs:
> (I don't have any autoload cookies).

Why don't you add some?

GuiltyDolphin:
Added some cookies, fixed the username/pkgname fields.

New output from `check-recipe.el`:

screenshot from 2016-01-11 08 49 11

GuiltyDolphin:
@dimitri 

Removed the non-used URL - should be all.

PR #2328, opened 2016-01-28
Update magit recipes

syohex:
- Update magit dependencies
- with-editor is hosted own repository now
- Add async recipe

See also
- https://github.com/milkypostman/melpa/pull/3542
- https://github.com/magit/magit/blob/b549e4e0e7d57954eef9011329a53641f3807dcf/lisp/magit.el#L19

### DIFF
--- a/recipes/magit.rcp
+++ b/recipes/magit.rcp
@@ -5,8 +5,7 @@
        :pkgname "magit/magit"
        :branch "master"
        :minimum-emacs-version "24.4"
-       :depends (dash)
-       :provide (with-editor)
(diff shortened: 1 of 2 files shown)


npostavs:
We already have async under the name emacs-async.

syohex:
Oh, I mistook.

syohex:
Thanks for reviewing. I have updated.

npostavs:
Merged, I just removed "add async recipe" from commit message.

The async vs emacs-async is a bit confusing, it might make sense to rename.

PR #2337, opened 2016-02-26
Updated matlab mode recipe as sources moved to git

csantosb:
For full discussion, refer to http://bit.ly/24t9fdp


npostavs:
Thanks, I removed also `:module` (which is CVS specific) and unshortened the URL, since git commit messages don't have a character limit.

PR #2338, opened 2016-03-03
fix jedi-core recipe (download raw, not HTML)

hmeine:


hmeine:
This is the first time I edited a recipe.  I also noticed that jedi-core depends on epc and python-environment.  Should I add these as dependencies?

npostavs:
> I also noticed that jedi-core depends on epc and python-environment. Should I add these as dependencies?

Yup, probably. If you have the recipe open `M-x el-get-auto-update-dependencies` should write the dependencies for you.

hmeine:
Interesting, thanks for the hint.  Unfortunately, the dependencies also contain auto-completion, which seems to reflect the truth, namely that I cannot use jedi-core with company-jedi, but ac will automatically be installed, activated and used.

npostavs:
> Unfortunately, the dependencies also contain auto-completion

Do you mean the dependencies written by `el-get-auto-update-dependencies`? I can't see why, it could be a bug in `el-get-auto-update-dependencies`.

hmeine:
Yes, I mean that.  I also see AC being installed and activated, without any .emacs entries or relevant customized variables.  When I tried el-get-remove auto-completion, I got complaints about two packages not being found.

For now, I gave up.  I spent too long trying to get this running, and don't have the time ATM.  Hope my feedback was at least better than nothing.

npostavs:
I've cherry picked this to `master` as 28c168434d17784943c0b229055c9fa77d083dd3, and added dependencies in 9bc149f1c715eb250f79527c53023c1ac3264374. Not really sure about the `auto-complete` thing, if you find time to investigate more, please open a new issue.

PR #2340, opened 2016-03-09
Feature/study

enzochiau:
Add the feature to merge "depends" keywords

### DIFF
--- a/el-get-bundle.el
+++ b/el-get-bundle.el
@@ -165,8 +165,7 @@
     (when (plist-member def :depends)
       (let ((old (el-get-as-list (plist-get def :depends))))
         (dolist (d old) (add-to-list 'ds d))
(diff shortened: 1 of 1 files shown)


npostavs:
Can you explain a bit what this change means? For example, how does it affect users?

enzochiau:
If I have some configure as below with keyword `depends`, it will not merge it with the original recipe. It ends up with only depends on `swiper` not `dash`, `with-editor` and `emacs-async` which presented in the original. 

    (el-get-bundle! magit
      :depends (swiper)
      :features (magit-gitflow)
      (require 'ivy)
      (setq magit-completing-read-function 'ivy-completing-read))

npostavs:
Thanks, I squashed your commits and updated the commit message with the explanation.

PR #2341, opened 2016-03-10
use github mirror for tablegen mode

vmiheer:
For now llvm viewvc is down let's use github in meantime.
@dholm, do you see any issue with this?

### DIFF
--- a/recipes/tablegen-mode.rcp
+++ b/recipes/tablegen-mode.rcp
@@ -2,8 +2,7 @@
        :description "Major mode for LLVM TableGen description files."
        :website "http://llvm.org/"
        :type http
-       :url "http://llvm.org/viewvc/llvm-project/llvm/trunk/utils/emacs/tablegen-mode.el?view=co"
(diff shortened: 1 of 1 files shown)


npostavs:
> For now llvm viewvc is down let's use github in meantime.

I was going to say it's silly to change the recipe just because a server goes down for a bit, but then I found https://llvm.org/bugs/show_bug.cgi?id=20381; apparently it's been down since 2014(?!!).

I think the `:website` should stay the same though, right? (setting it to `github.com` doesn't make sense at least)

vmiheer:
Hi @npostavs,
Updated change.
> setting it to github.com doesn't make sense at least

makes sense.
Wondering is there way to have multiple recipes for package?

npostavs:
> Wondering is there way to have multiple recipes for package?

No. I think there was some discussion about it in one of the issues here, but in general adding this would probably be difficult, and not very useful for most packages.

Issue #2335, opened 2016-02-19
dependency conflict (elpa & el-get) 


romankrv:
Hi all.
I use package and el-get together. Problem is: when I add magit via el-get I have conflict with dependency in package and el-get. Both use with-editor. How to make a separately use a with-editor package? How to separate the impact from other repositories?

thanks

![screen shot 2016-02-20 at 1 27 50 am](https://cloud.githubusercontent.com/assets/42733/13192087/47e2aa64-d771-11e5-9c63-4981ca4fe1cd.png)

npostavs:
Does it work to put `el-get` before `package-initialize`?

romankrv:
My config has a first a package system as step of the initialization after that run el-get.

romankrv:
you offer me to change the order of the initialization el-get as first?

npostavs:
> you offer me to change the order of the initialization el-get as first?

Yes.

npostavs:
Closing, as I don't see a bug here.

Call `(el-get)` before `(package-initialize)` to get el-get versions of the packages loaded, or else add `(:name dash :type elpa)` and `(:name with-editor :type elpa)` to `el-get-sources` in order to get elpa versions of the packages.

PR #2342, opened 2016-03-11
Add el-get-ignore-packages variable

enzochiau:
* This variable will define the packages which will not be installed by
  el-get. Which should solved the same package is installed by both
  el-get and other ways issue.

This should solve the issue like [#2335](https://github.com/dimitri/el-get/issues/2335)

### DIFF
--- a/el-get.el
+++ b/el-get.el
@@ -201,6 +201,9 @@
 (defvar el-get-next-packages nil
   "List of packages to install next, used when dealing with dependencies.")
 
+(defvar el-get-ignore-packages nil
+  "List of packages to be ignored.")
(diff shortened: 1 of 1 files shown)


npostavs:
Instead of adding a new variable, you can simply add to el-get-sources: e.g. `(add-to-list 'el-get-sources '(:name package-installed-via-melpa :type elpa))`, or `(add-to-list 'el-get-sources '(:name package-installed-via-some-other-way :type builtin))`

enzochiau:
That's cool. I don't know you can do the trick with `:type`. It works.

PR #2347, opened 2016-03-19
Add bats-mode recipe.

nevsan:


npostavs:
Shouldn't it be `:type github`?

nevsan:
It could be. The GitHub page advises using MELPA to install, but it could just as easily come from GitHub. Is it better practice in el-get to prefer the GitHub source over the MELPA package?

npostavs:
Well, it's certainly more common practice. And you can always do `M-x el-get-elpa-build-local-recipes` to get the `:type elpa` version, so it doesn't add that much.

nevsan:
Got it. Updated.

PR #2352, opened 2016-03-23
Fix broken link for bind-key recipe.

hikouki:
Fix broken link for bind-key recipe :sushi:

### DIFF
--- a/recipes/bind-key.rcp
+++ b/recipes/bind-key.rcp
@@ -1,5 +1,0 @@
-(:name bind-key
-       :description "Simplify key binding calls in elisp"
-       :type github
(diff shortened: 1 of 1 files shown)


npostavs:
Hmm, I wonder if there is any point in having separate bind-key and use-package recipes at this point.

hikouki:
Hi @npostavs 
Thank you for reply :beer: 

Sorry.. I didn't notice that there is a use-package recipe.
I think bind-key recipe is unnecessary, so I deleted it.
Is this OK?

npostavs:
Yeah, since the old bind-key recipe was broken anyway, I guess no one will miss it.

PR #2361, opened 2016-04-07
Update dic-lookup-w3m.rcp

dlwe:
url has changed


npostavs:
Picked to `master`. It looks like your repository is very far behind, I suggest you update.

PR #2362, opened 2016-04-09
Add airline-themes.

xuhdev:


npostavs:
Is the `:features` needed?

xuhdev:
I'm not sure, but maybe...

PR #2365, opened 2016-04-13
use github mirror for llvm mode

vmiheer:
For now llvm viewvc is down let's use github in meantime.
@npostavs, @dholm will you please review and merge?


npostavs:
Hah, I thought I reviewed and merged this one already, but I see that was #2341.

PR #2371, opened 2016-04-18
Fix Recipes for Rust Development

hadenpike:
The file extension .rs is now automatically associated with rust-mode. Racer is the code-completion mechanism for Rust. The Emacs integration between Rust-mode and rust-racer was split into the existing recipe emacs-racer.rcp, but the recipes did not reflect this. Finally, add dependencies so that emacs-racer runs without errors.

### DIFF
--- a/recipes/emacs-racer.rcp
+++ b/recipes/emacs-racer.rcp
@@ -2,4 +2,4 @@
        :description "Racer support for Emacs"
        :type github
        :pkgname "racer-rust/emacs-racer"
-       :depends (rust-mode company-mode s))
+       :depends (rust-mode company-mode dash s))
--- a/recipes/rust-racer.rcp
+++ b/recipes/rust-racer.rcp
@@ -6,7 +6,5 @@
        :pkgname "phildawes/racer"
        :description "Rust code completion and code navigation"
        :build '(("cargo" "build" "--release"))
-       :depends (rust-mode company-mode)
-       :load-path "editors/emacs"
-       :prepare (add-hook 'rust-mode-hook #'racer-activate)
-       :post-init (setq racer-cmd (concat default-directory "/target/release/racer")))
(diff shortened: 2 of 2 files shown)


recipes/emacs-racer.rcp
```diff
@@ -2,4 +2,5 @@
        :description "Racer support for Emacs"
        :type github
        :pkgname "racer-rust/emacs-racer"
-       :depends (rust-mode company-mode))
+       :depends (rust-mode rust-racer company-mode dash s)
```
npostavs: Seems `s` (but not `rust-racer`) was already added by #2370, can you rebase to take account of this, please?

recipes/rust-mode.rcp
```diff
@@ -1,4 +1,6 @@
 (:name rust-mode
        :type github
        :pkgname "rust-lang/rust-mode"
-       :description "Emacs mode for Rust")
+       :description "Emacs mode for Rust"
+       :prepare (autoload 'rust-mode "rust-mode" nil t)
+       :post-init (add-to-list 'auto-mode-alist '("\\.rs\\'" . rust-mode)))
```
npostavs: Generally, `auto-mode-alist` modification should go in `:prepare`, except that I see `;;;###autoload` cookies in rust-mode.el ([1], [2]), so neither the `:prepare` nor `:post-init` should be required.

[1]: https://github.com/rust-lang/rust-mode/blob/ed0b390e11be6642085ae05a57aa9ceae63211c4/rust-mode.el#L1280
[2]: https://github.com/rust-lang/rust-mode/blob/ed0b390e11be6642085ae05a57aa9ceae63211c4/rust-mode.el#L1329

hadenpike: Done.
Haden Pike
Computer Science
University of Kentucky
Class of 2016

PR #2373, opened 2016-04-25
Added agda-input recipe

nyraghu:
Agda input is an input method for easily typing mathematical and other Unicode characters.  The output of the recipe checker is attached as a text file.

[agda-input.txt](https://github.com/dimitri/el-get/files/235457/agda-input.txt)


npostavs:
> The output of the recipe checker is attached as a text file.

Just the output from `M-x el-get-check-recipe` should be enough (and short enough to paste inline).

nyraghu:
> Just the output from M-x el-get-check-recipe should be enough (and short enough to paste inline).

Oops, when I saw the instruction in the el-get manual to paste the recipe checker result in the pull request comment, I misunderstood it to mean the output of test-recipe.sh, and not check-recipe.sh or el-get-check-recipe.  Will do the right thing next time!

npostavs:
Maybe we need to clarify that page. Though most people don't bother reading it at all, since I seldom see any output from checker or tester posted!

nyraghu:
> Just the output from M-x el-get-check-recipe should be enough
I'm curious if the submitted recipes are tested with test-recipe.sh by travis.

npostavs:
> I'm curious if the submitted recipes are tested with test-recipe.sh by travis.

No, just check-recipe with some checks disabled (due to existing recipes failing them). See `.travis.yml` and `test/travis-ci.sh`.

PR #2374, opened 2016-04-25
Added proof-general recipe

nyraghu:
ProofGeneral.rcp in the current distribution of el-get packages an old version of Proof General from http://proofgeneral.inf.ed.ac.uk/releases/ProofGeneral-4.2.tgz

Proof General has since moved to GitHub, see http://coq-club.inria.narkive.com/Dp5A2NzL/proof-general-has-moved-to-github#post1

The recipe I'm submitting packages the GitHub version of Proof General.  I've named it proof-general.rcp to avoid a clash with the current ProofGeneral.rcp.  The output of the recipe checker is attached as a text file.

[proof-general.txt](https://github.com/dimitri/el-get/files/235236/proof-general.txt)

### DIFF
--- a/recipes/proof-general.rcp
+++ b/recipes/proof-general.rcp
@@ -1,0 +1,11 @@
+(:name proof-general
+       :description "A generic Emacs interface for interactive proof assistants."
+       :type github
+       :pkgname "ProofGeneral/PG"
+       :build `(("make" "clean")
+                ("make" ,(concat "EMACS=" el-get-emacs) "compile")
+                ("makeinfo" "doc/ProofGeneral.texi" "-o" "doc")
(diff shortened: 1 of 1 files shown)


npostavs:
> The recipe I'm submitting packages the GitHub version of Proof General. I've named it proof-general.rcp to avoid a clash with the current ProofGeneral.rcp.

Is there any point in keeping the old one?

nyraghu:
The announcement at the second URL I'd mentioned says that development of Proof General will now take place at GitHub, so it doesn't seem necessary now to keep the old recipe.

npostavs:
I removed the old one also: fb1124715c1e1acd25f1010894f300581fd8ec52

PR #2376, opened 2016-04-27
rcp/es mode

vedang:
- Adds a new recipe, `spark`, which is a es-mode dependency
- Updates the `es-mode` recipe and adds the `:depends` property

### DIFF
--- a/recipes/es-mode.rcp
+++ b/recipes/es-mode.rcp
@@ -3,4 +3,4 @@
        :description "An Emacs major mode for editing Elasticsearch requests"
        :type github
        :pkgname "dakrone/es-mode"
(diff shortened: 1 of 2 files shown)


npostavs:
Can we drop the `:features` from `es-mode` too? Looks like es-mode.el has `;;;###autoloads`: https://github.com/dakrone/es-mode/blob/4.1.1/es-mode.el#L780-L781
https://github.com/dakrone/es-mode/blob/4.1.1/es-mode.el#L758-L759

vedang:
Will do, I'll push a new patchset dropping `:features` for `es-mode`

npostavs:
Thanks.

PR #2378, opened 2016-04-28
Added hcl-mode recipe and dependency to get terraform-mode working

pcurry:
Terraform-mode installed without a dependency, so I added in the hcl-mode by the same author.


npostavs:
Does the terraform-mode recipe need to be updated?

pcurry:
It did need to be updated.  I missed it in my `git add` on the first commit.  Here it is.

PR #2389, opened 2016-05-18
emacs-racer: add rust-racer as a dependency.

hadenpike:


recipes/emacs-racer.rcp
```diff
@@ -2,4 +2,5 @@
        :description "Racer support for Emacs"
        :type github
        :pkgname "racer-rust/emacs-racer"
-       :depends (rust-mode company-mode dash s))
+       :depends (rust-mode rust-racer company-mode dash s)
+       :post-init (add-hook 'rust-mode-hook #'racer-mode))
```
npostavs: The hook should be added in `:prepare`, I think.

hadenpike: Done.

PR #2399, opened 2016-06-19
Update nim mode recipe and add commenter recipe

yuutayamada:
The nim-mode's repository was moved to nim-lang from reactormonk's and
the nim-mode is depending on commenter package, so I just added the recipe as well.
(https://github.com/nim-lang/Nim/wiki/editor-support)


yuutayamada:
I added missing :description for commenter package.

PR #2400, opened 2016-06-20
Fixed jedi-core and company-jedi recipes

### LINKED ISSUE #2350: error with dependencies in jedi-core.rcp
From [Melpa recipe](https://github.com/melpa/melpa/blob/master/recipes/jedi-core) we need to download few files
```lisp
files ("jedi-core.el" "jediepcserver.py" "Makefile" "setup.py"))
```
but in [your recipe](https://github.com/dimitri/el-get/blob/master/recipes/jedi-core.rcp) it\`s download only .el file
Could you fix this error? jedi:install-server doesn\`t work without this files

Crandel:
fixes #2350.

### DIFF
--- a/recipes/company-jedi.rcp
+++ b/recipes/company-jedi.rcp
@@ -2,4 +2,5 @@
        :description "Company backend for Python jedi."
        :website "https://github.com/syohex/emacs-company-jedi"
        :type github
-       :pkgname "syohex/emacs-company-jedi")
+       :pkgname "syohex/emacs-company-jedi"
+       :depends (jedi-core))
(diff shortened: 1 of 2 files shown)


npostavs:
So the only significant difference between jedi and jedi-core after this would be that jedi depends on auto-complete, and jedi-core doesn't, right?

I think you should add `:compile "jedi-core.el"` so that el-get won't try to compile jedi.el (it has `(require 'auto-complete)` which would cause compilation failure if you don't have that installed, I think.

Crandel:
Thank you for helping, It`s my first public pull request: )

npostavs:
Okay, I think if you can just add to the description field an explanation about the difference between jedi-core and jedi recipes it will be good to go.

npostavs:
Squashed to `master`, thanks for working on this :)

PR #2407, opened 2016-06-24
Recipe: add nim-emacs-module

yuutayamada:


recipes/nim-emacs-module.rcp
```diff
@@ -0,0 +1,8 @@
+(:name nim-emacs-module
+       :description "Make Emacs functions by Nim"
+       :type github
+       :pkgname "yuutayamada/nim-emacs-module"
+       :minimum-emacs-version "25"
+       :prepare
+       (eval-after-load "nim-mode"
+         '(el-get 'sync 'nim-emacs-module)))
```
npostavs: What's this `:prepare` block trying to accomplish? Trying to install `nim-emacs-module` when loading `nim-mode`? (but it only takes effect after `num-emacs-module` is installed...)

yuutayamada: Sorry, when I added this configuration, I encountered something like
"Cannot open load file ..." to use an autoloaded function in this
package. (and I added this configuration to avoid the error)
But I re-updated after I removed the :prepare section of
nim-emacs-module recipe, then the problem was gone... (maybe my emacs'
configuration problem)

PR #2408, opened 2016-06-24
Recipe: add flycheck-package

yuutayamada:


recipes/flycheck-package.rcp
```diff
@@ -0,0 +1,10 @@
+(:name flycheck-package
+       :type github
+       :pkgname "purcell/flycheck-package"
+       :depends (flycheck)
+       :minimum-emacs-version "24"
+       :prepare
+       (eval-after-load "flycheck"
+         '(progn
+            (el-get 'sync 'flycheck-package)
```
npostavs: So like the `nim-emacs-module` package, I guess `(el-get 'sync 'flycheck-package)`  makes no sense.

yuutayamada: Yes, you are right. I removed that `(el-get 'sync ...)`.

PR #2414, opened 2016-07-08
Add docker-tramp recipe

Yuki-Inoue:
The definition is taken from that of MELPA:
    https://melpa.org/#/docker-tramp

Note this is included in the dependency of recipe/docker.rcp .

### DIFF
--- a/recipes/docker-tramp.rcp
+++ b/recipes/docker-tramp.rcp
@@ -1,0 +1,4 @@
+(:name docker-tramp
+       :description "TRAMP integration for docker containers"
+       :type github
+       :pkgname "emacs-pe/docker-tramp.el" )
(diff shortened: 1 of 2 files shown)


npostavs:
> Note this is included in the dependency of recipe/docker.rcp .

"*is* included" or "*should* be" included? I don't see it there now.

Yuki-Inoue:
@npostavs Oh yeah,,, it "is not included" now but "should be included".  I'll modify the docker.rcp too...

Yuki-Inoue:
@Silex Adding docker-tramp to el-get.

Silex:
:+1: already the case in docker.el

``` elisp
;; Package-Requires: ((emacs "24.4") (dash "2.12.1") (docker-tramp "0.1") (magit-popup "2.6.0") (s "1.11.0") (tablist "0.70"))
```

PR #2423, opened 2016-07-26
Improve restclient recipes

Crandel:


recipes/know-your-http-well.rcp
```diff
@@ -2,4 +2,6 @@
        :description "HTTP headers, media-types, methods, relations and status codes, all summarized and linking to their specification."
        :website "https://github.com/for-GET/know-your-http-well"
        :type github
-       :pkgname "for-GET/know-your-http-well")
+       :pkgname "for-GET/know-your-http-well"
+       :load-path ("emacs")
+       :load ("emacs/know-your-http-well.el"))
```
npostavs: I see that there are `;;;###autoload` cookies, so the explicit load shouldn't be necessary.

Crandel: There are no *.el files in root directory, so nothing load from there, except explicit load from recipe

npostavs: The `:load-path` that you added looks correct, I just don't see why you need the `:load`, everything should be autoloaded.

Crandel: Fixed

PR #2437, opened 2016-09-01
update helm-bibtex and add ivy-bibtex

ericdanan:
There is now an additional package ivy-bibtex in the
tmalbsburg/helm-bibtex repository (and on MELPA). Helm-bibtex.el and ivy-bibtex.el are now two alternative
backends for the backend-independent bibtex-completion.el. The former
backend depends on helm whereas the latter depends on ivy (swiper).

So I updated the helm-bibtex and added the ivy-bibtex recipes, using the current
descriptions and dependencies.

Note that ideally, one should exclude the helm-bibtex.el file from the
ivy-bibtex package and the ivy-bibtex.el file from the helm-bibtex
package (as is done on MELPA). But I didn't find how to do this with el-get.

cc @tmalsburg, the author of helm-bibtex and ivy-bibtex.

### DIFF
--- a/recipes/ivy-bibtex.rcp
+++ b/recipes/ivy-bibtex.rcp
@@ -1,0 +1,5 @@
+(:name helm-bibtex
+       :type github
+       :description "A BibTeX bibliography manager based on Ivy."
+       :pkgname "tmalsburg/helm-bibtex"
+       :depends (swiper parsebib s dash f cl-lib biblio))
--- a/recipes/ivy-bibtex.rcp
+++ b/recipes/ivy-bibtex.rcp
@@ -1,5 +1,6 @@
-(:name helm-bibtex
+(:name ivy-bibtex
        :type github
        :description "A BibTeX bibliography manager based on Ivy."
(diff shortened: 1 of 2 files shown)


npostavs:
> Note that ideally, one should exclude the helm-bibtex.el file from the
> ivy-bibtex package and the ivy-bibtex.el file from the helm-bibtex
> package (as is done on MELPA). But I didn't find how to do this with el-get.

Yeah, there is no way, since el-get clones the git repository and git doesn't support partial repos (unlike say, SVN). But it might be worth adding a `:compile` directive and leaving out the file that doesn't apply for the package, I think it might avoid some compile warnings/errors.

recipes/ivy-bibtex.rcp
```diff
@@ -0,0 +1,5 @@
+(:name helm-bibtex
```
npostavs: The package `:name` should be the same as the file name (the automated test should catch this, not sure why it didn't, I can't access the log).

ericdanan:
Thanks, I added the `:compile` property and fixed the name.

Issue #2438, opened 2016-09-02
Cannot open load file "el-get-core"


bebekim:
I use emacs 24.5.1 on ubuntu 16.04
and when I open up emacs, I get a message saying problem with loading el-get-core

Using `--debug-init` I get the following message: 
```
Debugger entered--Lisp error: (file-error "Cannot open load file" "no such file or directory" "el-get-core")
  require(el-get-core)
  byte-code("\300\301!\210\300\302!\210\300\303!\207" [require cl el-get-core autoload] 2)
  load("/home/yhk/.emacs.d/elpa/el-get-5.1/el-get-autoloads" nil t)
  #[0 "\302\303\304\305\300\"\301\"\306\307#\207" [el-get "/home/yhk/.emacs.d/elpa/el-get-5.1" load expand-file-name format "%s-autoloads" nil t] 5 "\n\n(fn)"]()
  funcall(#[0 "\302\303\304\305\300\"\301\"\306\307#\207" [el-get "/home/yhk/.emacs.d/elpa/el-get-5.1" load expand-file-name format "%s-autoloads" nil t] 5 "\n\n(fn)"])
  package-activate-1([cl-struct-package-desc el-get (5 1) "Manage the external elisp bits and pieces you depend upon" nil nil nil "/home/yhk/.emacs.d/elpa/el-get-5.1" ((:keywords "emacs" "package" "elisp" "install" "elpa" "git" "git-svn" "bzr" "cvs" "svn" "darcs" "hg" "apt-get" "fink" "pacman" "http" "http-tar" "emacswiki") (:url . "http://www.emacswiki.org/emacs/el-get")) nil])
  package-activate(el-get)
  package-initialize()
  eval-buffer(#<buffer  *load*> nil "/home/yhk/.emacs.d/init.el" nil t)  ; Reading at buffer position 434
  load-with-code-conversion("/home/yhk/.emacs.d/init.el" "/home/yhk/.emacs.d/init.el" t t)
  load("/home/yhk/.emacs.d/init" t t)
[...]

PR #2439, opened 2016-09-02
[RFC] Is it time to switch 'el-get-allow-insecure to nil

manandbytes:
...and cut a new release? Plese, do not merge yet as, in the first place, I'm looking for feedback if I'm doing something completely stupid  ;-)


el-get-methods.el
```diff
@@ -23,16 +26,35 @@
...
+                                 (string-blank-p checksum)
+                               (let ((inhibit-changing-match-data t))
+                                 (string-match "\\`[ \t\n\r]*\\'" checksum)))))
+         (protocol (url-type (url-generic-parse-url url)))
+         (file-local (and (string= "file" protocol)
+                          (string= "" (url-host (url-generic-parse-url url))))))
+    (when (and (not el-get-allow-insecure)
+               (not (stringp (car (member protocol el-get-secure-protocols))))
+               (not file-local)
+               (not (string-match "^[-_\.A-Za-z0-9]+@" url)))
```
manandbytes: I can't remember a signle case of such URL and going to remove support for URLs starting with 'USERNAME@'.

npostavs: ssh+git URLs can start with `USERNAME@`, see https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols#The-SSH-Protocol. Apparently `ssh://USERNAME@...` also works, but github always offers `git@github.com...` for ssh URLs.

el-get-methods.el
```diff
@@ -14,6 +14,7 @@
 (require 'el-get-core)
 (unless (version< emacs-version "24.4")
   (require 'subr-x))
+(require 'url-parse)
```
npostavs: There might be some problems in relying on `url-parse`, see #1940.

test/el-get-tests.el
```diff
@@ -125,3 +125,49 @@ Following variables are bound to temporal values:
...
+                        "bzr+ssh://example.com/"
+                        "sftp://example.com/"
+                        "John.Doe-123_@example.com"))
+
+(ert-deftest el-get-insecure-check-secure ()
+  "Secure URL for a package without :checksum doesn't matter"
+  (dolist (url secure-urls)
+    (let ((el-get-allow-insecure nil)
+          (el-get-sources '((:name "dummy" :type github))))
+      (should-not (el-get-insecure-check "dummy" url)))))
```
npostavs: No need for the `should-not` here, returning `nil` is not really an important part of `el-get-insecure-check` that needs to be tested.

npostavs:
> ...and cut a new release?

We're currently in a "rolling release" model, which simplifies the questions of how to upgrade from an older release to a new one :) So before cutting a release we'd have to think about how to answer those.

el-get-custom.el
```diff
@@ -612,15 +612,15 @@ platforms where this recipe should apply"
...
-(defcustom el-get-allow-insecure t
+(defcustom el-get-allow-insecure nil
   "Allow packages to be installed over insecure connections."
   :group 'el-get
   :type 'boolean)
 
 (defcustom el-get-secure-protocols '("https" "ssh" "git+ssh" "bzr+ssh" "sftp")
   "List of secure protocols"
   :group 'el-get
+  :package-version '(el-get-custom . "6")
```
manandbytes: No sure if `el-get-custom` is the right package as `el-get-methods.el` actually depends on `el-get-secure-protocols`. Or even `(el-get . "6")`?

npostavs: The package should be just `el-get`, I think. `el-get-custom` is not versioned independently.

npostavs:
Hmm, perhaps we should tag some kind of release, since it seems that el-get is in melpa-stable right now (see #2438).

el-get-methods.el
```diff
@@ -24,26 +24,38 @@
                "methods"
                (file-name-directory (or load-file-name byte-compile-current-file buffer-file-name)))))
 
-(defun el-get-insecure-check (package url)
-  (let* ((checksum (plist-get (el-get-package-def package) :checksum))
+(defun el-get-insecure-check (PACKAGE URL)
+  "Check if it's safe to install PACKAGE from url.
```
manandbytes: Still not sure about the docstring...

manandbytes:
As it turned out this PR's main feature is defcustom for the list of secure URL protocols, I'm not sure what to do next. I'm going to create a separate PR for this, is it :ok:?

These commits are not ready anyway and should not block the rest:
- [RFC] Time to switch 'el-get-allow-insecure to nil
- Semantic URL parsing 
- [WIP] Don't consider URL starting with 'username@' as secure

el-get-methods.el
npostavs: Should be `from URL`, and I think the docstring should mention that an error is thrown if it's insecure/not safe. Otherwise looks good to me.

npostavs:
> I'm going to create a separate PR for this, is it :ok:?

Sure thing.

manandbytes:
>>    I'm going to create a separate PR for this, is it :ok:?

> Sure thing.

Done, #2441.

PR #2440, opened 2016-09-04
modify org-mode git url

rexshihaoren:
The old org-mode url would result in "fatal: read error: Connection reset by peer". After changed the git url to "http://orgmode.org/org-mode.git", it's now working. I have tested the new one with "test/test-recipe.sh org-mode".


npostavs:
Not sure we should do this, org's website lists the `git://` url as the one to use. That said, I'm seeing the `git://` one failing here too. It could be a temporary thing though?

rexshihaoren:
Ok, I have emailed its maintainer and will find out what happened.

rexshihaoren:
It's now working

PR #2441, opened 2016-09-04
Make list of secure protocols customizable

manandbytes:
Via 'el-get-secure-protocols, user can customize list of URL protocols
considered by 'el-get-insecure-check as secure.

By default, these protocols are considered as secure:
- https
- ssh
- git+ssh
- bzr+ssh
- sftp

### DIFF
--- a/el-get-methods.el
+++ b/el-get-methods.el
@@ -24,6 +24,7 @@
 
 (defun el-get-insecure-check (package url)
   (when (and (not el-get-allow-insecure)
+             (not (string-match "^file:///" url))
              (not (string-match "^https://" url))
              (not (string-match "^[-_\.A-Za-z0-9]+@" url))
              (not (string-match "^ssh" url)))
--- a/el-get-methods.el
+++ b/el-get-methods.el
@@ -24,26 +24,38 @@
                "methods"
                (file-name-directory (or load-file-name byte-compile-current-file buffer-file-name)))))
 
-(defun el-get-insecure-check (package url)
-  (let* ((checksum (plist-get (el-get-package-def package) :checksum))
+(defun el-get-insecure-check (PACKAGE URL)
+  "Raise an error if it's not safe to install PACKAGE from URL.
+
+When `el-get-allow-insecure' is non-nil, check if either of the
+following is true and retun nil:
+
+- URL's protocol is in `el-get-secure-protocols'
+
+- URL starts with 'file:///' (without hostname), so it points to the
+  local file
+
+- URL starts with username, i.e. 'username@example.com', also known as
+  SCP-like syntax
+
+- PACKAGE definition has a non-empty :checksum"
+  (let* ((checksum (plist-get (el-get-package-def PACKAGE) :checksum))
          (checksum-empty (or (not (stringp checksum))
                              (if (fboundp 'string-blank-p)
                                  (string-blank-p checksum)
                                (string-match-p "\\`[ \t\n\r]*\\'" checksum)))))
     (when (and (not el-get-allow-insecure)
-               (not (string-match "^file:///" url))
-               (not (string-match "^https://" url))
-               (not (string-match "^[-_\.A-Za-z0-9]+@" url))
-               (not (string-match "^sftp://" url))
-               (not (string-match "^bzr\\+ssh://" url))
-               (not (string-match "^git\\+ssh://" url))
-               (not (string-match "^ssh" url)))
+               (not (string-match "\\`file:///" URL))
+               (not (car (member 0 (mapcar (lambda (secure-proto)
+                                             (let ((proto-rx (concat "\\`" (regexp-quote secure-proto) "://")))
+                                               (string-match-p proto-rx URL))) el-get-secure-protocols))))
+               (not (string-match "\\`[-_\.A-Za-z0-9]+@" URL)))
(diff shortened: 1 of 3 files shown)


manandbytes:
Please, either rebase or merge instead of squashing into single commit ;-)

el-get-methods.el
```diff
@@ -22,17 +24,39 @@
...
+                                 (string-blank-p checksum)
+                               (string-match-p "\\`[ \t\n\r]*\\'" checksum)))))
+    (when (and (not el-get-allow-insecure)
+               (not (string-match "^file:///" URL))
+               (not (car (member 0 (mapcar (lambda (secure-proto)
+                                             (let ((proto-rx (concat "^" (regexp-quote secure-proto) "://")))
+                                               (string-match-p proto-rx URL))) el-get-secure-protocols))))
+               (not (string-match "^[-_\.A-Za-z0-9]+@" URL)))
+      ;; With not empty :checksum, we can rely on `el-get-post-install' for
+      ;; security.
```
npostavs: I originally wrote the comment claiming that we can rely on `el-get-post-install`, but I don't see a check directly in there. So the comment should be clarified (I guess I meant to rely on the functions that `el-get-post-install` calls?). And do we really need to check for a blank checksum? Wouldn't the normal check for checksum equality take care of that as well?

manandbytes: > And do we really need to check for a blank checksum? Wouldn't the normal check for checksum equality take care of that as well?

May be this is too defensive, but I think that empty `:checksum` should be considered here as if it is missing. Ideally, I would like to `'el-get-package-def` to be responsible for returning semantically meaningful values, and return nil for empty `:checksum`. I just see no use case for empty string.

manandbytes: > I originally wrote the comment claiming that we can rely on el-get-post-install, but I don't see a check directly in there. So the comment should be clarified (I guess I meant to rely on the functions that el-get-post-install calls?)

I'm not quite sure I follow what you are trying to say. What should I do? ;-)

npostavs: > I just see no use case for empty string.

I agree there is no use case for it, but there is no use case for a lot of other different non-checksum values. I don't see a need to specifically to look out for that one in particular. But  since you are writing the code, I'll leave the final decision on this up to you.

---

>> I originally wrote the comment claiming that we can rely on el-get-post-install, but I don't see a check directly in there. So the comment should be clarified (I guess I meant to rely on the functions that el-get-post-install calls?)
> 
> I'm not quite sure I follow what you are trying to say. What should I do? ;-)

Hmm, I was bit confused while writing that. I was looking to confirm that empty strings would be rejected correctly, and for some reason failed to find the `el-get-verify-checksum` call in `el-get-post-install`. So basically, the comment should also mention `el-get-verify-checksum` to help silly readers like present-me, and not assume that it's obvious like past-me did ;-)

manandbytes: > I agree there is no use case for it, but there is no use case for a lot of other different non-checksum values. But since you are writing the code, I'll leave the final decision on this up to you.

Why call methods deeper in the chain if we know right here that is should be noop anyway? Why give the chance for (potential) errors caused by empty checksum? If there is no use for the code/option, it shouldn't be there ;-)
So, It makes sense to start cleaning things up then, step by step.

> So basically, the comment should also mention

Just to double check: the comment or the docstring?

npostavs: >> So basically, the comment should also mention
> 
> Just to double check: the comment or the docstring?

The comment. No need to go into implementation details in the docstring.

el-get-methods.el
```diff
@@ -23,18 +25,23 @@
...
+                             (if (fboundp 'string-blank-p)
+                                 (string-blank-p checksum)
+                               (string-match-p "\\`[ \t\n\r]*\\'" checksum)))))
+    (when (and (not el-get-allow-insecure)
+               (not (string-match "^file:///" url))
+               (not (string-match "^https://" url))
+               (not (string-match "^[-_\.A-Za-z0-9]+@" url))
+               (not (string-match "^ssh" url)))
+      ;; With not empty :checksum, we can rely on `el-get-post-install' calling
+      ;; `el-get-verify-checksum' for security.
```
manandbytes: Hope it is ok.

el-get-methods.el
```diff
@@ -22,17 +24,39 @@
...
+         (checksum-empty (or (not (stringp checksum))
+                             (if (fboundp 'string-blank-p)
+                                 (string-blank-p checksum)
+                               (string-match-p "\\`[ \t\n\r]*\\'" checksum)))))
+    (when (and (not el-get-allow-insecure)
+               (not (string-match "^file:///" URL))
+               (not (car (member 0 (mapcar (lambda (secure-proto)
+                                             (let ((proto-rx (concat "^" (regexp-quote secure-proto) "://")))
+                                               (string-match-p proto-rx URL))) el-get-secure-protocols))))
+               (not (string-match "^[-_\.A-Za-z0-9]+@" URL)))
```
npostavs: These three regex test should use  ``` \` ``` instead of `^`.

manandbytes:
Any updates?

npostavs:
Merged, thanks.

> Any updates?

I think I'm supposed to get a notification when you push commits, but somehow I'm not seeing any (which is to say, you should add a message after pushing, otherwise I might not notice).

PR #2446, opened 2016-09-08
new recipe

tsypa:
A collection of Yasnippets for writing ECMAScript2015 in Emacs

### DIFF
--- a/recipes/es6-snippets.rcp
+++ b/recipes/es6-snippets.rcp
@@ -1,0 +1,5 @@
+(:name es6-snippets
+       :description  "A collection of Yasnippets for writing ECMAScript2015 in Emacs"
+       :type github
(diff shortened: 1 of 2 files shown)


recipes/es6-snippets.rcp
```diff
@@ -0,0 +1,5 @@
+(:name es6-snippets
+       :description  "A collection of Yasnippets for writing ECMAScript2015 in Emacs"
+       :type github
+       :pkgname "/CodyReichert/es6-snippets"
```
npostavs: There shouldn't be a leading slash here, I think (though maybe doubled slashes in URLs get ignored so it doesn't really matter?)

tsypa: yes, an extra slash

PR #2453, opened 2016-09-12
Allow the user to chose between sync/async el-get.

zhenya1007:
Prior to that change el-get bundle forced the `sync` argument to el-get to `'sync` (and then set  it to `nil` in some circumstances where `'sync` wouldn't work.  This change gives user the option to more finely control whether el-get is being run synchronously or asynchronously by the `el-get-bundle` macro.

### DIFF
--- a/el-get-bundle.el
+++ b/el-get-bundle.el
@@ -253,9 +253,14 @@ same.  If you wish to `require' more than one feature, then use
 `:features' property in FORM.
 
 The initialization FORM may start with a property list that
-describes a local recipe.  The FORM after the property list is
-treated as initialization code, which is actually an `:after'
-property of the local recipe.
+describes a local recipe.  The property list may include the keyword
+`:bundle-sync' with a value of either `t' or `nil' to request that
+`el-get-bundle' invoke `el-get' synchronously (respectively asynchronously).
+The keyword `:bundle-async' is the inverse of `:bundle-sync'.
+\(Note that the request to run el-get synchronously may not be respected in all
+circumstances: see the definition of `el-get-bundle-el-get' for details.)
+The FORM after the property list is treated as initialization code,
+which is actually an `:after' property of the local recipe.
 
 A copy of the initialization code is stored in a directory
 specified by `el-get-bundle-init-directory' and its byte-compiled
--- a/el-get-bundle.el
+++ b/el-get-bundle.el
@@ -23,6 +23,16 @@
   :type 'boolean
   :group 'el-get-bundle)
 
+(defcustom el-get-bundle-sync t
+  "t means to run el-get with the argument 'sync.
+The default can be overriden on a per-package basis by adding one of
+`:bundle-async [t|nil]'
+`:bundle-sync [t|nil]
+to the list of keywords that follow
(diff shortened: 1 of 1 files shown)


zhenya1007:
I know it's a bit lame to claim this, but I am struggling to see how my changes would have caused the CI test failure that is being reported.

npostavs:
I think the problem is that you used `cl-case`, which is not available on older Emacs.

el-get-bundle.el
```diff
@@ -23,6 +23,16 @@
   :type 'boolean
   :group 'el-get-bundle)
 
+(defcustom el-get-bundle-sync t
+  "t means to run el-get with the argument 'sync.
+The default can be overriden on a per-package basis by adding one of
+`:bundle-async [t|nil]'
+`:bundle-sync [t|nil]
```
npostavs: I think it's better to use just `:async`, `:sync` as the keywords: the `:bundle-` prefix seems redundant since it's going to occur within an `el-get-bundle` call anyway.

el-get-bundle.el
```diff
@@ -282,7 +295,7 @@ version is used if `el-get-bundle-byte-compile' is non-nil."
     ;; init script
     (setq src (plist-put src :after form))
 
-    `(el-get-bundle-el-get ',src)))
+    `(el-get-bundle-el-get ',src ,(and sync ''sync))))
```
npostavs: Not sure why you're using `and` here.

el-get-bundle.el
zhenya1007: I disagree.  The bundle-async/bundle-sync keywords are processed by the el-get-bundle macro; the rest of the keywords are processed by el-get.  Adding a `bundle-` prefix avoids a "name space clash" between the two sets of keywords.

el-get-bundle.el
zhenya1007: Seeing as how this is a "one-armed if", I think "when" might be the easiest to read, actually.

el-get-bundle.el
npostavs: Oh right, forgot about that, never mind then.

el-get-bundle.el
npostavs: I think moving one of the `quote`s outside of the `,` would help too.

zhenya1007: Except that, when the (let-bound) variable `sync` is nil (false), the macro would then expand to `(el-get aRecipe 'nil)`.  The `'nil` part is harmless, but it reads funny since `nil` is self-quoting (that was the reason I didn't do the `',` originally).

zhenya1007:
Okay, I think 'nil is worth it for the improvement in readability of the macro.  Thank you for pointing out that `cl-case` is not friendly to older Emacsen.  Let me know if you have other concerns; I am also happy to squash and re-submit the pull request, if that's your preference.

npostavs:
Code looks good now, only thing missing is a mention of the new keywords in the `el-get-bundle` docstring.

> I am also happy to squash and re-submit the pull request, if that's your preference.

No need since Github now provides a "Squash and merge" button.

zhenya1007:
>  only thing missing is a mention of the new keywords in the el-get-bundle docstring.

Good point.

PR #2458, opened 2016-09-13
remove emacs-w3m dependency from emms recipe

pasja:
* upstream dropped last.fm support
* w3m was only required by that part


pasja:
Sorry, this is clearly not enough, i will clear up properly the the dependencies in the :build section,
and update this pr

pasja:
I have add one more commit.
I was able to successfully install emms with that recipe.

PR #2461, opened 2016-09-16
Delegate install and update actions for github to git

manandbytes:
...for recipe without :url.


npostavs:
Are you saying update currently fails in some way? I just tried now (with `anaphora.rcp`) and it seems to work fine.

manandbytes:
> Are you saying update currently fails in some way

Yes, when `el-get-allow-insecure` is `nil`, see setup in `'el-get-github-update` test.
Test for `'el-get-git-pull` will follow later.

methods/el-get-github.el
```diff
@@ -78,11 +78,13 @@ USERNAME and REPONAME are strings."
                         el-get-github-default-url-type))))
     (el-get-github-url-private url-type username reponame)))
 
-(defun el-get-github-clone (package url post-install-fun)
+(defun el-get-github-clone (package _ post-install-fun)
```
npostavs: I think it's better to use `_url` so it's clear what parameter is being ignored.

methods/el-get-github.el
```diff
@@ -78,11 +78,13 @@ USERNAME and REPONAME are strings."
...
   "Clone the given package from Github following the URL."
-  (let ((url (or url (el-get-github-url package))))
-    (el-get-insecure-check package url)
-    (el-get-git-clone package url post-install-fun)))
+  (el-get-git-clone package (el-get-github-url package) post-install-fun))
+
+(defun el-get-github-pull (package _ post-install-fun)
```
npostavs: Same here.

test/el-get-tests.el
```diff
@@ -177,3 +177,20 @@ John.Doe-123_@example.com"))
             (el-get-sources '((:name "dummy" :type github :checksum checksum))))
         ;; TODO check for error message?
         (should-error (el-get-insecure-check "dummy" url) :type 'error)))))
+
+(ert-deftest el-get-github-update ()
+  "Update :type github"
+  (let* ((github-pkgname "user/dummy.el")
+         (el-get-sources `(,`(:name "dummy" :type github :pkgname ,github-pkgname)))
```
npostavs: A single backquote should suffice:
```elisp
`((:name ... ,github-pkgname))
```

npostavs:
> Yes, when el-get-allow-insecure is nil

Ah, would be good to mention this condition in the commit message.

manandbytes:
I'd like to shift focus of this PR to just `:type github` delegate `:install` and `:update` actions to `:type git` with a proper GitHub URL and going to:
- squash https://github.com/dimitri/el-get/pull/2461/commits/4a4baf15fbb18e1d7a8662b9c7ab69feb36e5889 and https://github.com/dimitri/el-get/pull/2461/commits/fe636235c6d3d05e0026b9f371215b836ca4fa59 into single commit
- followed by 
https://github.com/dimitri/el-get/pull/2461/commits/2a607f798678ffc890f7aa30b4d9ad76f0a4d8e5

Problem with `'el-get-git-pull` throwing an error when a) `el-get-allow-insecure` is nil and  b) URL is empty is a separate issue that deserves a separate PR.

WDYT?

npostavs:
Sure, sounds fine.

manandbytes:
> Sure, sounds fine.

Done, not sure why Emacs 25 still failing.

test/el-get-tests.el
```diff
@@ -177,3 +177,20 @@ John.Doe-123_@example.com"))
...
+
+(ert-deftest el-get-github-update ()
+  "Update :type github"
+  (let* ((github-pkgname "user/dummy.el")
+         (el-get-sources `((:name "dummy" :type github :pkgname ,github-pkgname)))
+         (el-get-allow-insecure nil))
+    (letf (((symbol-function 'el-get-package-is-installed)
+               (lambda (package)
+                 (string= package "dummy")))
+              ((symbol-function 'el-get-git-pull)
```
npostavs: The bindings after the first are indented too far.

npostavs:
> Done, not sure why Emacs 25 still failing.

Yeah, it's weird, I opened #2463 to try and see what's going on.

test/el-get-tests.el
npostavs: Also the `lambda` value itself is indented too far.

manandbytes: Done.

npostavs:
> Yeah, it's weird, I opened #2463 to try and see what's going on.

Okay, I think I figured it out, it seems that in Emacs 25, a failed `cl-assert` will try to enter the debugger directly when `debug-on-error` is non-nil (see `cl--assertion-failed`).

However, there is an actual problem hiding behind this, which is that `:type elpa` packages fail to install with a nil `el-get-allow-insecure`, just like `:type github` ones (the test is expected to fail for another reason, which is why it seems okay on other Emacs versions).

manandbytes:
> :type elpa packages fail to install with a nil el-get-allow-insecure

...because of `string-match`ing against nil url in

    (and (not el-get-allow-insecure)
         (not (string-match "^https://" url))
    ...)

which is here since acdcb6e.

manandbytes:
Ok, what's next?

npostavs:
Right, we can take care of the elpa problem separately.

PR #2468, opened 2016-09-26
Default :url to pass assertion in security check

### LINKED ISSUE #2467: Can't install ELPA packages without a "secure" :url
Steps to reproduce:

1. `emacs -Q`
2. Put this in `*scratch*`:
    
    ```elisp
    (setq user-emacs-directory "/tmp/egb/emacs")
    (url-retrieve
     "https://raw.githubusercontent.com/dimitri/el-get/master/el-get-install.el"
     (lambda (s)
       (goto-char (point-max))
       (eval-print-last-sexp)))
    (el-get-install "queue")
    ```
3. `M-x eval-buffer`

Expected result: `queue` is installed from ELPA.

Observed result: Installation fails with the message "apply: URL is nil, can’t decide if it’s safe to install package ’queue’".

Tested against GNU Emacs 25.1.1 (from https://emacsformacosx.com/) on OS X 10.11.6.

Additional comments: I believe this will happen for any ELPA recipe which does not have a `:url` that starts with `https` because of the `(el-get-insecure-check package url)` call at the top of `el-get-elpa-install`.  As far as I know, ELPA recipes ignore `:url` anyway, so I suspect this check is incorrect.  If we want to check if an ELPA recipe is "secure" then maybe we need to find out what URL `package-install` will use for ELPA?

To reiterate, this problem does affect other packages.  In addition to `queue` I've also encountered this with `adaptive-wrap` and `csv-mode`.

yurizhykin:
Not sure if this is acceptable solution, but since `url` is ignored for `elpa` method, it can be defaulted to something like "https://insecure", as for `emacswiki` for example. Fixes #2467.

### DIFF
--- a/methods/el-get-elpa.el
+++ b/methods/el-get-elpa.el
@@ -151,8 +151,13 @@ the recipe, then return nil."
                                    (bound-and-true-p package-archive-base)))
          ;; Prepend elpa-repo to `package-archives' for new package.el
          (package-archives (append (when elpa-repo (list elpa-repo))
-                                   (when (boundp 'package-archives) package-archives))))
-    (el-get-insecure-check package url)
+                                   (when (boundp 'package-archives) package-archives)))
+         ;; URL to use in security check
+         (package-url (or url (cdr (or elpa-repo
+                                       (assoc (package-desc-archive
+                                               (cadr (assq package package-archive-contents)))
+                                              package-archives))))))
+    (el-get-insecure-check package package-url)
 
     (unless (and elpa-dir (file-directory-p elpa-dir))
       ;; package-install does these only for interactive calls


npostavs:
I think we should check the package's archive URL, which should be:

```elisp
(cdr (or elpa-repo
         (assoc (package-desc-archive (cadr (assq package package-archive-contents)))
                package-archives)))
```

npostavs:
(24.3 and 23.4 test logs)
> ```methods/el-get-elpa.el:323:1:Error: the function `package-desc-archive' is not known to be defined.```

Ah, we need a different expression for older package.el versions.

yurizhykin:
Yep, I thought it was changed like `package-desc-doc` etc, so I tried to google old (pre 24.4) name, but it seems like there was no such function.

yurizhykin:
Will try to solve this in a couple of hours.

dsedivec:
> I think we should check the package's archive URL, which should be:
> 
> ```elisp
> (cdr (or elpa-repo
>          (assoc (package-desc-archive (cadr (assq package package-archive-contents)))
>                 package-archives)))
> ```

Thanks!  Just a heads up: I packaged this into advice around el-get-elpa-install and it seems that I got passed a string for the `PACKAGE` parameter, rather than a symbol, which caused `(assq package package-archive-contents)` to return `nil`.  This happened when installing moinmoin-mode required screen-lines.  I changed that to `(assq (el-get-as-symbol package) package-archive-contents)`.

In case anyone is interested, here's the advice I ended up with (which, as stated above, still only works in 24.5):

```elisp
(defun my:el-get-secure-url-patch (orig-fun package url &rest args)
  (apply orig-fun
         package
         (cdr (or (el-get-elpa-package-repo package)
                  (assoc (package-desc-archive
                          (cadr (assq (el-get-as-symbol package)
                                      package-archive-contents)))
                         package-archives)))
         args))

(advice-add 'el-get-elpa-install :around #'my:el-get-secure-url-patch)
```

(I made advice rather than patching el-get so I can keep upgrading el-get.)

npostavs:
> In case anyone is interested, here's the advice I ended up with (which, as stated above, still only works in 24.5):

Note this fails if `package-archive-contents` hasn't been initialized yet. https://github.com/dimitri/el-get/issues/2474#issuecomment-252452350

npostavs:
Okay, turns out there is already `package-archive-base` which we can use, though of course it's not entirely compatible across versions, see #2477.

PR #2472, opened 2016-09-29
Move iedit and livescript-mode origins

orivej:
emacswiki points from https://www.emacswiki.org/emacs/Iedit to https://github.com/victorhge/iedit . The version at github has bugfixes and improvements.

Neither version of livescript mode has been updated in two years, and there are no (known) further forks, but the version at https://github.com/piotrklibert/livescript-mode is one valuable commit ahead of https://github.com/bdowning/livescript-mode (it recognizes hyphenated names as single tokens).

### DIFF
--- a/recipes/iedit.rcp
+++ b/recipes/iedit.rcp
@@ -1,4 +1,7 @@
 (:name iedit
        :description "Edit multiple regions with the same content simultaneously."
-       :type emacswiki
+       :type github
+       :pkgname "victorhge/iedit"
+       ;; We load iedit so that it inserts iedit-toggle-key-default into the
+       ;; global key map.
        :features iedit)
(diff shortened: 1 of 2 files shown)


npostavs:
Since you're updating `iedit.rcp` anyway, is it possible to drop the `:features iedit`? I see that is has some `;;;###autoload` cookies.

orivej:
Unfortunately, without `features` `iedit-toggle-key-default` will not be inserted into the global key map: https://github.com/victorhge/iedit/blob/master/iedit.el#L226

npostavs:
Okay, can you just add a comment to the recipe saying that?

orivej:
Done.

npostavs:
Thanks, merged.

PR #2473, opened 2016-09-30
Consider :emacswiki over HTTPS as secure

manandbytes:


npostavs:
I think wiki recipes were specifically supposed to be considered insecure, because they can be edited anonymously.

manandbytes:
It fails in Emacs 25 only when comparing string `"URL is nil, can't decide if it's safe to install package 'dummy'"` due to the ASCII vs UTF quote characters:

expected:

	preferred charset: ascii (ASCII (ISO646 IRV))
    code point in charset: 0x27

actual:

	preferred charset: unicode (Unicode (ISO10646))
    code point in charset: 0x2019

Not sure how to fix this. Any hints?

npostavs:
> Not sure how to fix this. Any hints?

Honestly, I would suggest not checking for a specific error message. It doesn't seem like a useful test. If you insist on it, let-binding `text-quoting-style` might help.

---

But first, I'm not sure that Emacswiki packages over https *should* be considered secure, AFAICT, #1856 which introduced `el-get-allow-insecure` specifically left it out:

> In most cases, we assume any connection is insecure unless the URL
starts with "https://", "$USERNAME@", or "ssh". There are a few
exceptions: I'm assuming all Emacswiki packages are insecure,

npostavs:
> In most cases, we assume any connection is insecure unless the URL
> starts with "https://", "$USERNAME@", or "ssh". There are a few
> exceptions: I'm assuming all Emacswiki packages are insecure,

I'm therefore closing this as wontfix.

PR #2475, opened 2016-10-05
new recipes ferret-mode, wolfram-mode, ox-pandoc.

yyr:


yyr:
@npostavs Since I have commit access to the repo,  can I directly push the commits which only add new recipes? would it cause any inconvenience to others?

npostavs:
> Since I have commit access to the repo, can I directly push the commits which only add new recipes? would it cause any inconvenience to others?

@dholm does that occasionally too. Assuming the recipes have been tested it should be fine.

yyr:
Thanks.

PR #2478, opened 2016-10-10
El-Get recipe for Company.

zilti:


syohex:
`company-mode` is already registered.

https://github.com/dimitri/el-get/blob/2cb6f9b758b4a51c5d4ebff6a788929a4a38148c/recipes/company-mode.rcp

PR #2479, opened 2016-10-10
Create flycheck-clojure.rcp

zilti:


syohex:
Missing `:load-path` attribute.

zilti:
Corrected.

PR #2488, opened 2016-10-19
AucTeX needs GNU Make, not UNIX Make.

zilti:
As the title says. This goes unnoticed on most GNU/Linux distributions since there, `make` is just a symlink to `gmake`, but on e.g. FreeBSD, `gmake` needs to be called explicitly.

### DIFF
--- a/recipes/auctex.rcp
+++ b/recipes/auctex.rcp
@@ -4,13 +4,14 @@
        :type git
        :module "auctex"
        :url "git://git.savannah.gnu.org/auctex.git"
-       :build `(("./autogen.sh")
+       :build `(("export" "MAKE=gmake")
+                ("./autogen.sh")
                 ("./configure"
                  "--without-texmf-dir"
                  "--with-packagelispdir=$(pwd)"
                  "--with-packagedatadir=$(pwd)"
                  ,(concat "--with-emacs=" el-get-emacs))
-                ("make"))
+                ("gmake"))
        :build/darwin `(("./autogen.sh")
                        ("./configure"
                         "--without-texmf-dir"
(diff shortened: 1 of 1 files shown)


yyr:
hmm, on my system (debian testing) there is no gmake command..

```sh
$ gmake 
bash: gmake: command not found
✗ raghava-note:~/
$ make --version
GNU Make 4.1
Built for x86_64-pc-linux-gnu
Copyright (C) 1988-2014 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
raghava-note:~/
$ 
```

zilti:
I see... Is there a `:build/bsd` or something?

syohex:
`:build/berkeley-unix` ? You can see it in some recipes.

zilti:
Ok, this should be ok now, then :)

recipes/auctex.rcp
```diff
@@ -4,13 +4,14 @@
        :type git
        :module "auctex"
        :url "git://git.savannah.gnu.org/auctex.git"
-       :build `(("./autogen.sh")
+       :build `(("export" "MAKE=gmake")
```
npostavs: I think calling `export` has no effect because each command runs in its own subshell.

zilti:
Would it be better to drop this pull request and create another "clean" one with only one commit, or is this ok?

npostavs:
Can you fix the indentation please?

> Would it be better to drop this pull request and create another "clean" one with only one commit, or is this ok?

No need for a new pull request, I'll just squash when it's ready.

PR #2497, opened 2016-11-15
Add writeroom-mode dependency

dsedivec:
writeroom-mode depends on visual-fill-column.  I also
removed ":shallow nil" because I cannot see why it needs to be
there, but feel free to leave it in if someone knows better than
me.

### DIFF
--- a/recipes/writeroom-mode.rcp
+++ b/recipes/writeroom-mode.rcp
@@ -2,4 +2,4 @@
        :type github
        :description "Writeroom-mode: distraction-free writing for Emacs."
        :pkgname "joostkremers/writeroom-mode"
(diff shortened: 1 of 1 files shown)


npostavs:
> I also
> removed ":shallow nil" because I cannot see why it needs to be
> there, but feel free to leave it in if someone knows better than
> me.

Can you test it with `:shallow t`? There are a few packages that don't like to be installed in a shallow repository (e.g., magit checks for tags to figure out its version, which will not be present in a shallow clone).

In this case, it seems to have been there since the recipe was added in #1125, @yyr: do you remember why you used `:shallow nil`?

PR #2499, opened 2016-11-20
Update auctex.rcp

zilti:
That's the only way to make it work reliably on BSD.

### DIFF
--- a/recipes/auctex.rcp
+++ b/recipes/auctex.rcp
@@ -11,7 +11,7 @@
                  "--with-packagedatadir=$(pwd)"
                  ,(concat "--with-emacs=" el-get-emacs))
                 ("make"))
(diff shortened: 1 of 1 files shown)


recipes/auctex.rcp
```diff
@@ -11,7 +11,8 @@
                  "--with-packagedatadir=$(pwd)"
                  ,(concat "--with-emacs=" el-get-emacs))
                 ("make"))
-       :build/berkeley-unix `(("./autogen.sh" "MAKE=gmake")
+       :build/berkeley-unix `(("sed -i'' -e 's/MAKE=make/MAKE=gmake/g' autogen.sh")
```
zilti: Unfortunaltely, setting an environment variable doesn't work.

PR #2502, opened 2016-11-30
Don't use builtin CEDET by default

npostavs:
@MHOOO wrote at https://github.com/dimitri/el-get/commit/85eca62e323303531cfd17b97b06f09e69446248#commitcomment-20000437

> Isnt el-get supposed to stay up-to-date, instead of using builtins?

There's not really a hard rule about it, but since CEDET has been builtin since 23.3 which is earlier than the earliest Emacs we support anyway, probably anyone installing it would want the git version. Somehow I didn't think of it when I merged that change.

> This broke my config, since cedet is no longer being loaded from git.

I will a wait bit before merging to see if anyone else has some comments about this, but you can add `(:name cedet :builtin nil)` to `el-get-sources` to fix this in the meantime.

---

```
Users installing CEDET via el-get probably wanted the git version.  For
those who want the builtin, they can add (:name cedet :builtin "23.3")
to `el-get-sources'.

Also move the :post-init block to :prepare so that the :lazy nil hack
becomes redundant.
```

### DIFF
--- a/recipes/cedet.rcp
+++ b/recipes/cedet.rcp
@@ -1,5 +1,7 @@
 (:name cedet
-       :builtin "23.3"
+       ;; While CEDET is builtin since Emacs 23.3, CEDET users
+       ;; probably want the latest from git.
+       ;; :builtin "23.3"
        :website "http://cedet.sourceforge.net/"
        :description "CEDET is a Collection of Emacs Development Environment Tools written with the end goal of creating an advanced development environment in Emacs."
        :type git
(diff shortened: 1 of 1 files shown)


xuhdev:
The reason I proposed to add a builtin keyword is that when I tried to install something depending on CEDET, el-get also tries to install the latest version for me which causes problems on older versions of Emacs. My guess is that CEDET has been too integrated into Emacs that non-latest Emacs versions may not be officially supported.

npostavs:
Hmm, so it sounds like using the builtin CEDET breaks some configs, and using the git head CEDET breaks others. I'm not sure what the default should be.

npostavs:
I am looking at CEDET's git repo, and it seems to have been last updated in 2019. So probably the Emacs' builtin version is just as good in terms of up-to-dateness.

PR #2509, opened 2016-12-31
mu4e.rcp: Use the autogen.sh script instead of autoreconf.

xuhdev:


npostavs:
It looks liks the autogen scripts calls configure too, so I guess that step should be removed from the recipe?

xuhdev:
@npostavs Updated, thanks!

PR #2512, opened 2017-01-22
Change savannah urls from git to http.

yyr:
Old urls are not responding any more. (may be only for me??)


npostavs:
I just tried `git clone git://git.sv.gnu.org/geiser.git` now and it worked fine.

yyr:
Hmm.. Might be some thing to do with my local network.

PR #2516, opened 2017-01-28
etags-select now uses emacsmirror type

alexg0:
emacsmirror type tends to be lot more reliable then emacswiki.

### DIFF
--- a/recipes/etags-select.rcp
+++ b/recipes/etags-select.rcp
@@ -1,5 +1,4 @@
 (:name etags-select
        :description "Select from multiple tags"
-       :type http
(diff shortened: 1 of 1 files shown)


npostavs:
Maybe it's better to use `:type emacswiki`? (which does take from `https://raw.github.com/emacsmirror/emacswiki.org` by default)

alexg0:
@npostavs I have no objection.  Is there a standard for a preferred approach?

PS.  also think there should be a global change for all emacswiki.org recipes to one of the github types.

npostavs:
> Is there a standard for a preferred approach?

Standard? Not really, it just seems logical to use `:type emacswiki` for stuff from emacswiki.

PR #2519, opened 2017-02-07
Add recipe helm-tramp

masasam:


masasam:
Thank you for reviewing.
I do not know about "Missing: depends" so please tell me concretely

npostavs:
Ah sorry, el-get doesn't (yet) grab dependencies from the header like package.el does, so it needs to be put into the recipe manually. Actually there is a semi automatic method, if you open the recipe file and do `M-x el-get-auto-update-dependencies` it should write in the correct `:depends` block for you.

masasam:
Thanks for your polite advice.
Fixed according to your advice.

PR #2520, opened 2017-02-13
recipe: evil-magit

agsdot:


syohex:
`evil-magit` is moved to https://github.com/emacs-evil/evil-magit . The recipe should be updated.

agsdot:
Just added a new commit to revise it. Thanks for the notification @syohex.

PR #2521, opened 2017-02-18
Replace missing dependency inflection with jump on clj-refactor

f0086:
"inflection" is no longer part of elpa and it is now found in
the "jump" package.

### DIFF
--- a/recipes/clj-refactor.rcp
+++ b/recipes/clj-refactor.rcp
@@ -2,5 +2,5 @@
        :description "A collection of simple clojure refactoring functions"
        :type github
        :depends (dash s clojure-mode yasnippet paredit multiple-cursors cider
-                      edn inflections hydra)
+                      edn jump hydra)
        :pkgname "magnars/clj-refactor.el")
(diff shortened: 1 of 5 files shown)


f0086:
Yes, the `inflections.rcp` exists, but as I said, it is no longer part of elpa.
I can install `jump` over el-get.

npostavs:
> Yes, the `inflections.rcp` exists, but as I said, it is no longer part of elpa.

Then we should remove that recipe, since it doesn't work.

> I can install `jump` over el-get.

Don't we need to add a recipe for that?

> `Debugger entered--Lisp error: (wrong-type-argument arrayp nil)`

Hmm, that error should be more user-friendly.

f0086:
I've added the jump package and removed the inflection package.
All dependencies are adjusted accordingly.

npostavs:
Thanks!

npostavs:
> > `Debugger entered--Lisp error: (wrong-type-argument arrayp nil)`
> 
> Hmm, that error should be more user-friendly.
> 

This should now say something like 
`Debugger entered--Lisp error: (error "Couln’t find package ‘inflections’")`

ddf128195bc8e1184ecc8804df9a80b5a90156e0

PR #2525, opened 2017-03-14
taskpaper-mode

yfwu:
TaskPaper major mode for emacs


npostavs:
Hmm, last update was in 2011, not sure we should add such abandoned projects.

npostavs:
> Hmm, last update was in 2011, not sure we should add such abandoned projects.

Oh, actually I see now it's being maintained again since 2016.

PR #2530, opened 2017-05-07
add recipe for helm-frame

chee:
opens your helm buffer in a dedicated frame in the middle of the screen, similar to rofi or Alfred.

screenshot of helm-frame working in sandbox
^ screenshot of it working in a sandbox with only `helm` and `helm-frame` installed


dholm:
@chee Thanks, merged!

PR #2531, opened 2017-05-08
helm.rcp: add :depends block (async)

SkydiveMike:
helm.el has (require 'async)


dholm:
@SkydiveMike Thanks, merged!

PR #2544, opened 2017-06-09
Don't generate autoloads for proof-general

### LINKED ISSUE #2543: proof-general fails to install
When I attempt to use el-get to install proof-general (PG), I see the following:

```
Error (el-get): while installing proof-general: Cannot open load file: no such file or directory, /home/.../generic/proof-site
```

Where `/home/...` is the path I happened to be in when running the installation. If I subsequently reload then it complains about trying to load it at the path `~/.emacs.d/el-get/generic/proof-site`. The path where the file is actually located is `~/.emacs.d/el-get/proof-general/generic/proof-site`.

As best I can tell, this seems to be a bizarre interaction between the [`pg-init.el` file](https://github.com/ProofGeneral/PG/blob/master/pg-init.el) and el-get. The file attempts to locate PG by its own path. For some reason I cannot entirely fathom, el-get takes the contents of `pg-init.el` and dumps them into `~/.emacs.d/el-get/.loaddefs.el`. Surprise, this is in a different directory and now the file doesn't load properly.

This behavior is almost entirely opaque to me so I'm not sure whether I should be reporting it here or to PG. If the problem is upstream please say so (ideally along with a concrete hint I can take to them so they know what to do to fix it).

Thanks.

npostavs:
Fixes #2543.
```
It has some autoloaded statements which depend on the file they are
evaluated from, which breaks when the destination autoload file is
'~/.emacs.d/el-get/.loaddefs.el'.  The existing
'generic/proof-site.el' file already provides sufficient autoload
definitions.
```

### DIFF
--- a/recipes/proof-general.rcp
+++ b/recipes/proof-general.rcp
@@ -7,5 +7,5 @@
                 ("makeinfo" "doc/ProofGeneral.texi" "-o" "doc")
                 ("makeinfo" "doc/PG-adapting.texi" "-o" "doc"))
        :info "doc"
-       :load "generic/proof-site.el"
+       :autoloads "generic/proof-site.el"
        :website "http://proofgeneral.inf.ed.ac.uk/")


elliottslaughter:
I can confirm that this fixes the bug. Can this be merged? Thanks.

npostavs:
Thanks for testing, merged.

PR #2555, opened 2017-07-27
Add spotify.el recipe

lorenzhs:


recipes/spotify.el.rcp
```diff
@@ -0,0 +1,4 @@
+(:name spotify.el
+       :type git
+       :url "https://github.com/remvee/spotify-el"
```
npostavs: Could you use `:type github` instead please?

lorenzhs: Of course, fixed it.

npostavs: Perfect, thanks.

PR #2558, opened 2017-08-11
Fix symlink creation bug regarding elpa packages (#2556)

### LINKED ISSUE #2556: Symlinks to elpa package not created during `el-get-update`
Hi,

For some time now, I had noticed that executing `el-get-update-all` would fail while updating certain packages with an error complaining that no directory with the corresponding package's name could be found inside `el-get-dir`. I noticed that this error happened only during the update of packages fetched from elpa (as I had to manually fix the symlinks to their installation directories).

While having a quick look inside `methods/el-get-elpa.el`, I realized that the function `el-get-elpa-update` which is called to update such packages, after removing and re-installing the package (using `el-get-elpa-install-package` instead of `el-get-elpa-install`), never calls `el-get-elpa-symlink-package` on non-windows operating systems, and thus the symlink to the installation directory never gets created anew.

As a temporary fix, I tried placing `el-get-elpa-update`'s call to `el-get-elpa-symlink-package` outside the conditional statement and everything works fine now.

Being new to this and all, I don't really know what is the proper procedure for reporting/fixing this bug beyond this point.

maxchaos:
Fixes #2556

### DIFF
--- a/methods/el-get-elpa.el
+++ b/methods/el-get-elpa.el
@@ -237,8 +237,8 @@ first time.")
     ;; in windows, we don't have real symlinks, so its better to remove
     ;; the directory and copy everything again
     (when (memq system-type '(ms-dos windows-nt))
-      (delete-directory (el-get-elpa-package-directory package) t)
-      (el-get-elpa-symlink-package package)))
+      (delete-directory (el-get-elpa-package-directory package) t))
+    (el-get-elpa-symlink-package package))
   (funcall post-update-fun package))
 
 (defun el-get-elpa-remove (package url post-remove-fun)


npostavs:
Won't it give an error because the old link is not removed?

maxchaos:
The old link inside `el-get-dir` gets removed by a previous call to `el-get-elpa-remove` inside `el-get-elpa-update` (line 233). As far as what the conditional does, on windows machines it would remove the package's actual installation directory (**not** the symlink), which makes no sense to me either way.

npostavs:
> As far as what the rest of the conditional does, on windows machines it would remove the package's actual installation directory (not the symlink), which makes no sense to me either way.

Hmm, right, it looks completely broken on Windows. Perhaps nobody uses this...

PR #2562, opened 2017-09-01
add magithub dependencies

ericdanan:
* Add missing magithub dependency ghub+
* Add recipes for ghub+ and its dependencies: apiwrap and ghub
* Change magithub recipe file type from .el to .rcp (the .el file worked though so perhaps this was on purpose on which case you may not want to merge this last commit)

@vermiculus and @tarsius FYI.


yyr:
Thanks for the pull, Merged.

PR #2565, opened 2017-09-06
Added all-the-icons

nanjj:


yyr:
Hi, I memoize was added to the repository in other commit, could you update pull request with only all-the-icons. 

Thank you.

PR #2566, opened 2017-09-06
Delete cirrus.rcp

midlis:
github.com/pidu/cirrus.git is not accessible any more.


yyr:
I wonder if there is any archive over the net we could use instead of deleting this.

yyr:
Ohh I just checked, the recipe was added by me and I don't use amazon services at all. So I think no body will miss the package if this is gone.

Merging.. Thanks

PR #2567, opened 2017-09-13
Make the Daredevil SKK updatable

elim:
Using a cp command instead of a mv command to enable re-execute the build step.


yyr:
Thanks

elim:
Thank you, too.

PR #2569, opened 2017-09-19
Add missing org-ref dependencies

GuiltyDolphin:
The org-ref recipe was missing some dependencies (which can be seen at http://melpa.milkbox.net/#/org-ref) which was preventing it from running correctly.

Note that it appears that `ivy` is called `swiper` in the el-get distribution (I assume these are installed together, seeing as they are in the same repository).


yyr:
Thank you

PR #2581, opened 2017-11-24
Add general, convenience wrappers for keybindings

agsdot:
Rather than using [evil-leader](https://github.com/cofi/evil-leader), some people use general.el for their keybindings in evil. 

And as an added bonus, [general](https://github.com/noctuid/general.el) can "provides a more convenient way to bind keys in emacs for **both** evil and non-evil users."


yyr:
Thank you

PR #2583, opened 2017-11-24
Add magit-rockstar

agsdot:
"commit like a rockstar" .... and .... provide "commands which manipulate author and committer dates"


yyr:
Thank you

PR #2605, opened 2018-01-16
Create method 'gitlab'.

gavk:


gavk:
It took the recipe 'mo-git-blame', but the repository moved to https://gitlab.com. First I wanted to change the recipe type to 'git', but I decided to do it like 'github'.

.travis.yml
```diff
@@ -14,5 +14,5 @@ script:
   - byte-compile pkg/*.el
   - byte-compile -Werror *.el methods/*.el
   - ert-tests
-  - check-recipes -Wno-features -Wno-github -Wno-autoloads recipes/
+  - check-recipes -Wno-features -Wno-github -Wno-gitlab -Wno-autoloads recipes/
```
npostavs: It looks like there's just six recipes using gitlab urls right now, maybe we should just change them all to `:type gitlab`?

gavk: Yes. In addition to the existing 6 recipes, there is also "mo-git-blame", which moved to gitlab.
type git => type gitlab
url "https://gitlab.com/warsaw/winring.git" => pkgname "warsaw/winring"

yyr:
I wonder why we need separate zip tar methods for gitlab? can't we just **only** clone the package from gitlab?

PR #2607, opened 2018-01-29
* recipes/ido-completing-read-plus.rcp: ido-ubiquitous subsumed into this

yyr:
old ido-ubiquitous package should to be removed with this. But lets keep for a while, since there is procedure for old users to get informed about this change.


yyr:
Unintentionally merged. but I believe its not a problem.

PR #2608, opened 2018-02-11
Fix mu4e could not find mu4e/mu4e.info

coldnight:
While I'm installing `mu4e` via el-get, I got an error that no `mu4e/mu4e.info` file, I have fixed that problem, please review.

### DIFF
--- a/recipes/mu4e.rcp
+++ b/recipes/mu4e.rcp
@@ -11,7 +11,9 @@
                           "mu"
                           (el-get-package-directory 'mu4e))))
        :build `(("./autogen.sh")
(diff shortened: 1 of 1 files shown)


coldnight:
That seems my fault, I can't reproduce this anymore. Sorry to bother you.

coldnight:
I know why, before this my `emacs` command is point to version 22 that installed by system, and there is no error about generate `mu4e/mu4e.info`. After I have made those changes in `mu4e`'s recipe, I got and error, so I make the `emacs` command to point to the newest version that I installed. Problem solved!

PR #2610, opened 2018-03-10
fix pdf-tools recipe build

pasja:
* until they fix it upstream
* https://github.com/politza/pdf-tools/issues/325


npostavs:
Huh, I see our automated tests missed the `org-grep-rcp` mistake because it only checks `.rcp` files.

PR #2611, opened 2018-03-13
Supplement missing dependency in the dockerfile-mode.rcp

elim:
The dockerfile-mode requires the "s" library.
ref: https://github.com/spotify/dockerfile-mode/commit/52f821c


elim:
thx!

Issue #2412, opened 2016-07-03
google code refrences still in recipie database.


timoc:
Hi,

I tried to install org-outlook (which depends on soap-client) and it failed. After some digging i came to the conclusion that the recipe was wrong because it pointed at the now defunct code.google.com. A quick grep of the recipe folder picks up other recipes. 
in the last days of google code, they offered a mechanism to 'transfer to github' so it is likley now where these packages are maintained.

`tim@zangbandv2[2]:~ $ grep -r code.google.com .dotfiles/zemacsen/el-get/el-get/recipes/`
- [X] ~~soap-client.rcp:       :website "http://code.google.com/p/emacs-soap-client/"~~
- [X] ~~soap-client.rcp:       :url "https://code.google.com/p/emacs-soap-client/")~~
- [ ] confluence-el.rcp:       :website "http://code.google.com/p/confluence-el"
- [X] ~~go-def.rcp:       :pkgname "code.google.com/p/rog-go/exp/cmd/godef"~~
- [X] ~~emacswiki/struct+.rcp::description "has moved to http://code.google.com/p/emacs-ontology-project/source"~~
- [X] ~~csharp-mode.rcp:       :website "https://code.google.com/p/csharpmode/"~~
- [ ] project-mode.rcp:       :website "http://code.google.com/p/emacs-project-mode/"
- [X] ~~geben.rcp:       :website "https://code.google.com/p/geben-on-emacs/"~~
- [ ] vc-clearcase.rcp:       :url "https://code.google.com/p/vc-clearcase/"
- [ ] xahk-mode.rcp: :url http://ergoemacs.googlecode.com/svn/trunk/packages/xahk-mode.el2 (#2634)

AndersonTorres:
Sites I have found. I will try to update the recipes and fix some issues (creating the github projects).

[ ] confluence-el: https://sourceforge.net/projects/confluence-el/
[ ] vc-clearcase https://github.com/alex-hhh/vc-clearcase
[ ] xahk-mode.el: https://github.com/xahlee/xahk-mode.el



[ ] emacs-project-mode https://code.google.com/archive/p/emacs-project-mode/source/default/source

PR #2616, opened 2018-04-09
Update geben recipe to use MELPA

martinburchell:
See https://github.com/dimitri/el-get/issues/2412

No longer available at code.google.com. Use MELPA as recommended by https://github.com/ahungry/geben


recipes/geben.rcp
```diff
@@ -1,6 +1,7 @@
 (:name geben
-       :website "https://code.google.com/p/geben-on-emacs/"
        :description "GEBEN is a software package that interfaces Emacs to DBGp protocol with which you can debug running scripts interactively."
-       :type svn
-       :url "http://geben-on-emacs.googlecode.com/svn/trunk/"
+       :website "https://github.com/ahungry/geben"
+       :features geben
+       :type elpa
+       :repo ("melpa" . "http://melpa.org/packages/")
```
npostavs: Usually, we prefer recipes that taken from the source (i.e., `:type github` in this case), otherwise you may as well just use `el-get-elpa-build-local-recipes`.

PR #2621, opened 2018-05-23
Add new package Orgalist from Elpa

daimrod:
This library provides Org mode's plain lists in non-Org buffers, as a
minor mode.


npostavs:
I think you should get this recipe automatically if you run `M-x el-get-elpa-build-local-recipes`

daimrod:
Indeed, `M-x el-get-elpa-build-local-recipes` did the work. Thanks.

PR #2625, opened 2018-06-10
Removed unnecessary dependencies from lispy.rcp

zilti:
See also https://github.com/melpa/melpa/blob/master/recipes/lispy.


npostavs:
MELPA installation uses the dependencies from the package's headers, el-get does not.

See https://github.com/abo-abo/lispy/blob/master/lispy-pkg.el

zilti:
Fixed, I forgot to clear my el-get dir before testing with a clean install.

zilti:
The dependencies in the list right now are those needed to install lispy and enable lispy-mode in a buffer with lisp code. I assume the other two might be needed for special features.

PR #2630, opened 2018-06-19
adding vue-mode to recipes

sorenmat:
adding vue-mode to recipes

### DIFF
--- a/recipes/vue-mode.rcp
+++ b/recipes/vue-mode.rcp
@@ -1,0 +1,5 @@
+(:name vue-mode
+       :description "Emacs major mode for vue.js"
+       :type github
+      ;; :depends cl-lib
+       :pkgname "AdamNiederer/vue-mode/vue-mode.el")


recipes/vue-mode.rcp
```diff
@@ -0,0 +1,5 @@
+(:name vue-mode
+       :description "Emacs major mode for vue.js"
+       :type github
+      ;; :depends cl-lib
+       :pkgname "AdamNiederer/vue-mode/vue-mode.el")
```
npostavs: The `:pkgname` is not correct.

recipes/vue-mode.rcp
```diff
@@ -0,0 +1,5 @@
+(:name vue-mode
+       :description "Emacs major mode for vue.js"
+       :type github
+      ;; :depends cl-lib
```
npostavs: The `Package-Requires: ` in [vue-mode.el](https://github.com/AdamNiederer/vue-mode/blob/0.4/vue-mode.el#L9) should be converted to `:depends` entries. You can try using `el-get-auto-dependencies` for this.

PR #2631, opened 2018-06-25
adding go-add-tags

sorenmat:
Add field tags for struct fields.


recipes/go-add-tags.rcp
```diff
@@ -0,0 +1,4 @@
+-(:name go-add-tags
+-       :description "Add field tags for struct fields. This package is inspired by GoAddTags of vim-go."
+-       :type github
+-       :pkgname "syohex/emacs-go-add-tags/go-add-tags.el")
```
alexott: `pkgname` should point to repo itself (`syohex/emacs-go-add-tags`), not to the file...

PR #2635, opened 2018-07-24
Added go-dlv.el

nanjj:


recipes/go-dlv.rcp
```diff
@@ -0,0 +1,6 @@
+(:name go-dlv
+       :type github
+       :description "Debug Go programs interactively with the GUD."
+       :pkgname "benma/go-dlv.el"
+       :depends (go-mode)
+       :features go-dlv)
```
npostavs: It looks like the upstream code uses `;;;###autoload` cookies, so the `:features` shouldn't be needed.

nanjj: Fixed

PR #2639, opened 2018-08-22
adding rg recipe

cbbrowne:
And apparently some others that have collected...


recipes/rg.rcp
```diff
@@ -0,0 +1,4 @@
+(:name rg
+       :description "Emacs Search based on rip-grep"
+       :type git
+       :url "https://github.com/dajva/rg.el.git")
```
npostavs: Please use `:type github`.

recipes/devel-notes.rcp
```diff
@@ -0,0 +1 @@
+(:name devel-notes :type emacswiki)
```
npostavs: This can probably be left out, since emacswiki recipes can be generated with `el-get-emacswiki-build-local-recipes`

recipes/shadchen.rcp
```diff
@@ -0,0 +1,5 @@
+(:name shadchen
+       :description "Pattern matching library"
+       :type git
+       :url "https://github.com/VincentToups/shadchen-el.git"
+       :features shadchen)
```
npostavs: Isn't this library dead?

PR #2644, opened 2018-09-21
Fix w32 elpa directory copy for Emacs 26.1

### LINKED ISSUE #2633: emacs 26 on windows doesn't install elpa packages successfully
The call to "copy-directory" on line 143 of methods/el-get-elpa.el needs to have " nil t" added to the end of it so that the whole parent directory tree is created when copying it.
Otherwise, the install fails deep inside make-directory with a no such file or directory error.

Perhaps this changed because a (possibly long-standing) bug was fixed in emacs 26 where it was always creating the parent hierarchy if it didn't exist, even if the "parents" flag wasn't passed..

npostavs:
Fixes #2633.
```
In Emacs 26, copy-directory copies the directory itself, based on
whether NEWNAME is directory name (ends with a "/").  Earlier Emacs
versions actually checked whether NEWNAME was existing directory, so
we happened to get the right behaviour most of the time.

Pass t for COPY-CONTENTS to ensure the correct behaviour all of the
time.
```

### DIFF
--- a/methods/el-get-elpa.el
+++ b/methods/el-get-elpa.el
@@ -140,8 +140,14 @@ the recipe, then return nil."
       (if (memq system-type '(ms-dos windows-nt))
           ;; in windows, we have to actually copy the directories,
           ;; since symlink is not exactly reliable on those systems
-          (copy-directory (el-get-elpa-package-directory package)
-                          (file-name-as-directory (expand-file-name package el-get-dir)))
+          (let ((arglist (list (el-get-elpa-package-directory package)
+                               (file-name-as-directory (expand-file-name package el-get-dir))
+                               nil nil t)))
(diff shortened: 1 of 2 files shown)


npostavs:
Hum, failing check seems be due to the test script not handling force pushes correctly:

```
The command "check-recipes -Wno-features -Wno-github -Wno-autoloads recipes/" exited with 0.
0.00s$ check-whitespace
fatal: Invalid symmetric difference expression 87bb3f9481077f67d6520800b9ce5b8bf88fde5c...2fd9c80c98c36a5517b94255ab2cc5c1b329aec0
The command "check-whitespace" exited with 128.
```

Everything else passes fine.

npostavs:
> Everything else passes fine.

Hmm, I somehow missed the compile error under 23.4 earlier.

PR #2647, opened 2018-09-24
Update vimpulse-surround raw GH source URL

TheDahv:
It appears GitHub have updated the URL structure to find raw file contents since this was last looked at.

### DIFF
--- a/recipes/vimpulse-surround.rcp
+++ b/recipes/vimpulse-surround.rcp
@@ -1,4 +1,4 @@
 (:name vimpulse-surround
        :description "Emulate surround.vim in Vimpulse"
        :type http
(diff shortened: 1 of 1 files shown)


TheDahv:
Hi, I'm new to Emacs and found this as I was trying to get up and running. Apologies if I missed anything, but it seems like a simple change and got things unblocked for me.

npostavs:
As far as I can tell, the official URL remains the same, it just redirects to the raw.githubusercontent.com one. So it's probably better to keep the current URL in case Github decides to change the redirection. And if redirection is broken for you, find out why (could it be an issue of https vs http?).

But for this particular recipe, it looks like it should just be changed to `:type github` and bypass the whole issue.

PR #2650, opened 2018-10-22
add new recipe

raoulhatterer:


raoulhatterer:
Attempt to add oef-mode recipe in el-get master branch

PR #2651, opened 2018-10-23
Raoulhatterer oef mode recipe

raoulhatterer:
close #2650


raoulhatterer:
third attempt to publish off-mode recipe in el-get/recipes

PR #2662, opened 2018-11-30
Add orgmode to el-get

zilti:
Latest stable and patched orgmode for Emacs.


zilti:
The problem that's left is that, just like with package.el, it should be loaded instead of the built-in orgmode, but I don't know how to make el-get do that.

zilti:
Ah, right! I'm gonna request a small change over there, then.

PR #2664, opened 2018-12-07
Updating org-mode.rcp

zilti:
Setting the branch to `maint`. This branch contains the latest stable version including patches. `master` is often buggy or even broken.


yyr:
one of the reason people use el-get to be always on bleeding edge of all packages, may be we could have another recipe org-mode-stable or something? 

I am not sure about this, any thoughts @npostavs

alexott:
was fixed in another PR

PR #2666, opened 2018-12-14
Add recipe for figlet.el

nanjj:


yyr:
thanks for the pull and package as well

PR #2672, opened 2018-12-26
Add forge recipe

ericdanan:
Cc @tarsius who just released the forge package

I started from the magit recipe and amended it as needed. I think the build step is only needed for the info manual but it would perhaps be safer to for @tarsius to confirm that everything is correct.


tarsius:
Yes. Compiling and the generation of the autoloads file can be left to the package manager.

yyr:
thanks for the recipe

PR #2673, opened 2019-01-01
Add missing dependencies for lsp-mode

stephan-cr:
Lsp-mode requires some dependencies. This patch adds the missing dependencies to the recipe.


yyr:
thanks

PR #2677, opened 2019-01-27
Recipe for doom themes and modeline

stephan-cr:
This PR adds recipes for:

- doom themes
- its corresponding modeline, doom-modeline, and
- shrink-path which is required by doom-modeline


yyr:
thank you

Issue #2482, opened 2016-10-10
Option to install last tag?


zilti:
Is there currently a way to have El-Get install the checkout with the latest tag, thus mimicking MELPA Stable? There's e.g. CIDER which I want to have stable releases only.

npostavs:
This is currently not supported. Patches welcome.

zilti:
I'm thinking about implementing a `:checkout 'latest-tag` which does that, would this be ok, or do you favour another way of denoting that?

npostavs:
Yeah, that sounds good. A possible variation would be `:checkout #'el-get-latest-tag` where `el-get-latest-tag` is the name of the function which returns the tag. This has the advantage that it's easy to substitute other functions with different policies (e.g., `el-get-latest-nonbeta-tag` or something), though it's a bit more verbose. If you implement this, you can decide which way you like best :)

zilti:
I went with your idea of adding a function. Though I'm not really sure to what file I should add this function?

npostavs:
`el-get-recipes.el` perhaps?

PR #2679, opened 2019-01-31
Added ability to use functions at git's :commit option

zilti:
Concerning issue #2482

Now, two questions: 

1. Is this a good approach?
2. Should the function take an argument that denotes the VCS used so this change can be added to others  like Mercurial, or are there better ways to figure out the VCS?

### DIFF
--- a/el-get-recipes.el
+++ b/el-get-recipes.el
@@ -324,6 +324,10 @@ Use this to modify environment variable such as $PATH or $PYTHONPATH."
                                   :test #'string= :from-end t)
                path-separator)))
 
+(defun el-get-latest-tag ()
+  "Returns the latest tag of the git repository of the given package."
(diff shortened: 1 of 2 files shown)


zilti:
This is still work in progress. Actually, `el-get-start-process-list` is giving me some headaches. What would the preferred way be to use a return value from one process in the list in the next?

zilti:
I think passing the name and the method would suffice, since if the recipe is needed, it can be fetched using `el-get-package-def`. 

I could modify `el-get-start-process-list` so that if an entry is a lambda instead of a list, that lambda gets passed the return value of the last process?

npostavs:
> I could modify `el-get-start-process-list` so that if an entry is a lambda instead of a list, that lambda gets passed the return value of the last process?

Hmm, a lambda is also a list, so this could get messy. How about starting with just a plain `call-process` and see about making asynch if needed later?

PR #2683, opened 2019-02-06
Create cider-decompile.el

zilti:


recipes/cider-decompile.el
```diff
@@ -0,0 +1,5 @@
+(:name cider-decompile
+       :description "An extension to CIDER which provides a decompilation command"
+       :type github
+       :pkgname "clojure-emacs/cider-decompile"
+       :depends (cider))
```
stephan-cr: The recipe should depend on [javap-mode](https://github.com/clojure-emacs/cider-decompile/blob/5d87035f3c3c14025e8f01c0c53d0ce2c8f56651/cider-decompile.el#L10).

stephan-cr:
Thank you!

PR #2685, opened 2019-02-06
Create evalator.rcp

zilti:


recipes/evalator.rcp
```diff
@@ -0,0 +1,4 @@
+(:name evalator
+       :description "Emacs package for interactive evaluation and transformation of data."
+       :type github
+       :pkgname "seanirby/evalator")
```
stephan-cr: The recipe should depend on [helm](https://github.com/seanirby/evalator/blob/f30da4da48c0b3f3cfa1fc1c7cfdb53ffe79df36/evalator.el#L9).

yyr:
@zilti could you update the recipe with helm as a dependency as suggested?

stephan-cr:
Thanks a lot!

PR #2690, opened 2019-03-01
Support latest magit update

nanjj:


recipes/magit.rcp
```diff
@@ -5,7 +5,7 @@
        :pkgname "magit/magit"
        :branch "master"
        :minimum-emacs-version "24.4"
-       :depends (dash emacs-async ghub let-alist magit-popup with-editor)
+       :depends (dash emacs-async ghub let-alist magit-popup with-editor transient)
```
ericdanan: Looking at `magit-pkg.el`, it seems to me that it should rather be
```
:depends (emacs-async dash git-commit lv transient with-editor)
```
But note that there is not yet an el-get recipe for the `lv` package.

recipes/magit.rcp
```diff
@@ -5,7 +5,7 @@
        :pkgname "magit/magit"
        :branch "master"
        :minimum-emacs-version "24.4"
```
ericdanan: Looking at `magit-pkg.el` the minimum emacs version seems to be `25.1`.

recipes/transient.rcp
```diff
@@ -0,0 +1,21 @@
+(:name transient
+       :website "https://github.com/magit/transient#readme"
+       :description "Transient commands used by magit."
+       :type github
+       :pkgname "magit/transient"
+       :branch "master"
+       :minimum-emacs-version "24.4"
```
ericdanan: Looking at `transient.el`, the minimum emacs version seems to be `25.1`.

recipes/transient.rcp
```diff
@@ -0,0 +1,21 @@
...
+       :depends (dash hydra)
```
ericdanan: Lookint at `transient.el`, it doesn't seems to me that `hydra` is required.

recipes/transient.rcp
```diff
@@ -0,0 +1,21 @@
...
+       :info "docs"
+       :load-path "lisp/"
+       :compile "lisp/"
+       ;; Use the Makefile to produce the info manual, el-get can
+       ;; handle compilation and autoloads on its own.  Create an
+       ;; empty autoloads file because transient.el explicitly checks for
+       ;; a file of that name.
+       :build `(("make" ,(format "EMACSBIN=%s" el-get-emacs) "info")
+                ("touch" "lisp/transient-autoloads.el"))
```
ericdanan: I didn't see where `transient.el` checks for a `transient-autoloads.el` file, so I'm not sure this line is necessary.

recipes/transient.rcp
```diff
@@ -0,0 +1,21 @@
...
+       :build/berkeley-unix `(("gmake" ,(format "EMACSBIN=%s" el-get-emacs) "info")
+                              ("touch" "lisp/transient-autoloads.el"))
```
ericdanan: I didn't see where `transient.el` checks for a `transient-autoloads.el` file, so I'm not sure this line is necessary.

recipes/transient.rcp
```diff
@@ -0,0 +1,21 @@
...
+       ;; assume windows lacks make and makeinfo
+       :build/windows-nt (with-temp-file "lisp/transient-autoloads.el" nil))
```
ericdanan: I didn't see where `transient.el` checks for a `transient-autoloads.el` file, so I'm not sure this line is necessary.

recipes/magit.rcp
ericdanan: Re. adding an `lv` el-get-recipe, the `lv.el` file is inside the `abo-abo/hydra` github repository. I don't know how to write a recipe that fetches only this file from the repo.

recipes/transient.rcp
tarsius: It was until a few days ago but not anymore.

recipes/magit.rcp
npostavs: > I don't know how to write a recipe that fetches only this file from the repo.

There's no way to do that exactly. It's possible to write it as an http recipe (via raw.githubusercontent.com), or else just depend on `hydra`.

ericdanan: OK, I imagine it's not a big deal to depend on hydra even if one only wants lv.

tarsius: As I said in the response to another inline comment (*), magit no longer depends on lv.

(*) It is my impression that replies to inline comments not being noticed is a common issue. I think it is because clicking on the notification about the reply often (always?) fails to take the user to the correct place within the pull-request

ericdanan: Sorry @tarsius , I did notice your reply but thought it was only about `transient`.  So this is also true for `magit`, even though `lv` is still listed as a dependency in `magit-pkg.el`?

tarsius: Yes. I forgot to remove the reference to lv from magit. I'll do that tomorrow.

ericdanan: Thanks.

Also, since `git-commit` is inside the `magit` repo, I think it doesn't need to be listed as dependency by `el-get`. So @nanjj I suppose it should be good with
```
:depends (emacs-async dash transient with-editor)
```

tarsius: Unless `el-get` uses `package.el` to install `magit` you also do not need `emacs-async`.

See https://github.com/jwiegley/emacs-async/blob/master/async-bytecomp.el#L28-L52 for why and how Magit uses this.

tarsius:
Does `el-get` normally create an autoloads file for packages that it installs? I think it should.

> Create an empty autoloads file because magit.el explicitly checks for a file of that name.

Please create a proper autoloads file instead. You can use `make loaddefs` for that.

ericdanan:
> Does el-get normally create an autoloads file for packages that it installs? I think it should.

It does but stores autoloads for all packages in a common file. If I understand correctly, `magit` specifically wants the file `magit-autoloads.el` to exist, so the `el-get` recipe creates an empty file with that name.

recipes/magit.rcp
ericdanan: To be sure, is it also normal that the `el-get` recipe for `with-editor` does not depend on `async` although `async` is listed as a dependency in `with-editor.el`? (and if so then isn't the `async` dependency unnecessary in `magit-pkg.el` as well?)

recipes/transient.rcp
nanjj: depends on lv, so I put hydra here.

nanjj:
fixed

ericdanan:
Tnanks @nanjj . You may have overlooked @tarsius comment that `transient` no longer depends on lv, so the hydra dependency can be dropped entirely.

tarsius:
And `with-editor` does not need `emacs-async`.

nanjj:
Got and Fixed

PR #2691, opened 2019-03-10
add a recipe: aweshell

yfwu:
Aweshell is a powerful extension to the original eshell.

### DIFF
--- a/recipes/aweshell.rcp
+++ b/recipes/aweshell.rcp
@@ -1,4 +1,6 @@
 (:name aweshell
        :type github
        :description "Awesome shell extension base on eshell with wonderful features!"
-       :pkgname "manateelazycat/aweshell")
+       :pkgname "manateelazycat/aweshell"
+       :post-init
+       (require 'aweshell))
--- a/recipes/aweshell.rcp
+++ b/recipes/aweshell.rcp
@@ -2,5 +2,5 @@
        :type github
        :description "Awesome shell extension base on eshell with wonderful features!"
        :pkgname "manateelazycat/aweshell"
-       :post-init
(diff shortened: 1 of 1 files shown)


recipes/aweshell.rcp
```diff
@@ -0,0 +1,6 @@
+(:name aweshell
+       :type github
+       :description "Awesome shell extension base on eshell with wonderful features!"
+       :pkgname "manateelazycat/aweshell"
+       :post-init
+       (require 'aweshell))
```
npostavs: Is this `require` really needed?

yfwu: The author doesn't use ``autoload`` in the code. I have no idea which function should I add ``autoload`` in the recipe, so I use ``require`` to load it. Any suggestion of fixing this?

npostavs: > I have no idea which function should I add `autoload` in the recipe,

The interactive commands are usually a good place to start. May ask the package author too?

yfwu:
This should work. When the user initializes the ``eshell-mode``, the ``aweshell`` will be required.

recipes/aweshell.rcp
```diff
@@ -0,0 +1,6 @@
+(:name aweshell
+       :type github
+       :description "Awesome shell extension base on eshell with wonderful features!"
+       :pkgname "manateelazycat/aweshell"
+       :prepare
+       (autoload 'eshell "aweshell" nil t))
```
npostavs: This looks strange and likely to break eshell to me. Maybe use `(eval-after-load 'eshell '(require 'aweshell))` instead?

PR #2692, opened 2019-03-11
Added company-lsp

nanjj:
company-lsp is the company completion backend for lsp-mode

### DIFF
--- a/recipes/company-lsp.rcp
+++ b/recipes/company-lsp.rcp
@@ -1,0 +1,7 @@
+(:name company-lsp
+       :description "Company completion backend for lsp-mode "
+       :type "github"
+       :pkgname "tigersoldier/company-lsp"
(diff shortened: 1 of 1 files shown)


recipes/company-lsp.rcp
```diff
@@ -0,0 +1,8 @@
+(:name company-lsp
+       :description "Company completion backend for lsp-mode "
+       :type "github"
+       :pkgname "tigersoldier/company-lsp"
+       :depends (dash cl-lib s seq company-mode lsp-mode)
+       :post-init (progn
+                    (require 'company-lsp)
```
npostavs: The `require` shouldn't be needed here, since `:post-init` happens after the feature is loaded.

nanjj: fixed

PR #2693, opened 2019-03-17
Update forge and magit-popup dependencies

ericdanan:
Cc @tarsius

Forge no longer needs magit-popup. Also, I take it from #2690 that magit-popup does not need emacs-async with el-get.


tarsius:
Looks correct.

PR #2705, opened 2019-05-13
Fixed issues in recipes/

AtlasJan:
Altered permissions on one file and cleaned up messy directory.


npostavs:
Thanks. When I merged the stylus-mode.rcp addition, I though the 3 dots in `.../stylus-mode.rcp` were a bit odd, but I didn't realize it was due to a strange file name. I kind of assumed the `check-recipes` test would have caught something like that; I'll have to investigate why it didn't.

PR #2706, opened 2019-05-14
Check names of added files for sanity

npostavs:
Trying to catch problems like the one fixed in #2705 ahead of time.

```
* test/travis-ci.sh (check-filenames): New function.
* .travis.yml: Use it.  Bump Emacs 26 to 26.2.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ env:
   - EMACS_VERSION=24.3
   - EMACS_VERSION=24.5
   - EMACS_VERSION=25.3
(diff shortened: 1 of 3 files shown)


npostavs:
It catches the error:

```
$ check-filenames
"recipes/(:name stylus-mode        :description \"Emacs major mode for stylus template highlighting.\"        :type github        :pkgname \"vladh/stylus-mode.rcp": File name does not consist of -+./_ or ASCII letters or digits.
The command "check-filenames" exited with 1.
```

I've merged the first commit in.

PR #2712, opened 2019-07-28
fix-github-rawurl

masarakki:


npostavs:
Does this need fixing? As far as I know, the `https://github.com/<user/repo>/raw` URLs redirect to the equivalent `https://raw.githubusercontent.com/...` one anyway. And if you look at github's file view, the "Raw" button uses the first form so I guess it is preferred.

masarakki:
on my computer, redirect not works.

```
Error (el-get): while installing ruby-end: Failed to find end of headers in HTTP response from https://github.com/rejeep/ruby-end.el/raw/master/ruby-end.el for package ruby-end; see buffer  *http github.com:443*
```

so i changed raw url.

And this commit contains more important fix that repository name changed from `ruby-end` to `ruby-end.el`

npostavs:
> on my computer, redirect not works.

Hmm, works here. Does `url-retrieve-synchronously` work for you (e.g., evaluate the expression below)?

```elisp
(display-buffer (url-retrieve-synchronously "https://github.com/rejeep/ruby-end/raw/master/ruby-end.el"))
```

> do you think it's better to fix only repository name?

Probably. Another possibility is to change the recipe to `:type github`, I don't really see why was defined as `:type http` in the first place.

masarakki:
> (display-buffer (url-retrieve-synchronously "https://github.com/rejeep/ruby-end.el/raw/master/ruby-end.el"))
say "Contacting  host: github.com:443" but not finish.
and it same with "https://github.com"

> (display-buffer (url-retrieve-synchronously "https://raw.githubusercontent.com/rejeep/ruby-end.el/master/ruby-end.el"))
it works.

I think this is not a problem of redirection, but here is not a place to talk about it.

I changed to fix only repo-name.

npostavs:
> I think this is not a problem of redirection, but here is not a place
> to talk about it.

I guess it's either you have some firewall getting in the way, or
maybe you've run into [Emacs
Bug#36810](https://debbugs.gnu.org/36810), see if `(setq
gnutls-algorithm-priority "NORMAL:-VERS-TLS1.3")` helps.

> I changed to fix only repo-name.

Thanks, merged.

masarakki:
thanks!

PR #2715, opened 2019-08-28
Remove graphql dependency in forge and ghub recipes

ericdanan:
Cc @tarsius


tarsius:
Yep, that's not needed anymore.

PR #2716, opened 2019-09-03
Add autoload and auto-mode-alist preparation

rlaboiss:
This commit alows this recipe to work out of the box. Otherwise users will have to add the settings to their personal configuration.


rlaboiss:
There was a typo in the file, sorry.

PR #2717, opened 2019-09-03
Add autoload and auto-mode-alist preparation

rlaboiss:
This commit alows this recipe to work out of the box. Otherwise users will have to add the settings to their personal configuration.


yyr:
ideally this should be autoload cookie in the source file. Merging anyway

rlaboiss:
> ideally this should be autoload cookie in the source file. Merging anyway

I agree. I will try to propose a pull request upstream. Thanks.

PR #2729, opened 2019-11-18
Create company-php.rcp

zilti:


recipes/company-php.rcp
```diff
@@ -0,0 +1,7 @@
+(:name company-php
+       :description "Company version of PHP auto completion source"
+       :type github
+       :pkgname "xcwen/ac-php"
+       :build (("rm" "-f" "ac-php.el")
+               ("sed" "-i" "s/\(require \\x27company\)/\(require \\x27company-mode\)/g" "company-php.el"))
```
alexott: Why we need these 2 commands in recipe? Why not submit PR to the original package to fix this problem there?

alexott (changes requested):

PR #2730, opened 2019-11-18
Update php-mode.rcp

zilti:
Fixing repository.


alexott (changes requested):
The package is now located at https://github.com/emacs-php/php-mode - there is an automatic redirect from `ejmr/php-mode` repository

npostavs:
#2971 has fixed this meanwhile.

PR #2736, opened 2019-11-22
Create phpunit.rcp

zilti:


recipes/phpunit.rcp
```diff
@@ -0,0 +1,5 @@
+(:name phpunit
+       :description "Emacs mode to phpunit and test command line tool"
+       :type github
+       :pkgname "nlamirault/phpunit.el"
+       :depends (php-mode))
```
alexott: The package lists following dependencies: `(s "1.12.0") (f "0.19.0") (pkg-info "0.6") (cl-lib "0.5") (emacs "24.3")` - I think that we need to add some of them explicitly

alexott (changes requested):

PR #2758, opened 2020-03-28
Get minimap-mode from upstream GitHub, as ELPA package is out of date

rrthomas:


rrthomas:
This fixes two particular problems: first, the `minimap-highlight-line` setting is respected, and secondly, the face used to highlight the line can be customized.

PR #2760, opened 2020-05-05
Use original up-to-date request.el for the recipe

shevchuk:
A newer version from original maintainer is still supported so I think it will be more natural to use.

abingham's request.el is 0.2.0 and I suppose it is abandoned
tkf's request.el is 0.3.2

Reason to change:
vuiet.el and lastfm.el and probably some others require latest capabilities of request.el

### DIFF
--- a/recipes/request.rcp
+++ b/recipes/request.rcp
@@ -2,6 +2,6 @@
        :description "Easy HTTP request for Emacs Lisp"
        :type github
        :submodule nil
-       :pkgname "abingham/emacs-request"
+       :pkgname "tkf/emacs-request"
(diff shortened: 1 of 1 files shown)


falloutphil:
Hi,

When I install request using `el-get-install`, the install succeeds, but the library isn't available - i.e. `(require 'request)` is not locating the package.  I've checked the MELPA install and that works OK so it appears to be something peculiar to el-get or my setup?

I note the repo has a makefile - is that being executed as part of the install?

Any ideas?  Can anyone confirm the el-get package is working for them?

PR #2764, opened 2020-05-26
Swiper now depends on avy

lugia-kun:
Now swiper seems to depend on avy (while `make`).

This change makes using avy installed via el-get, but is slightly hacky.

### DIFF
--- a/recipes/swiper.rcp
+++ b/recipes/swiper.rcp
@@ -1,10 +1,10 @@
 (:name swiper
        :description "Gives you an overview as you search for a regex."
        :type github
-       :depends (cl-lib)
+       :depends (cl-lib avy)
        :pkgname "abo-abo/swiper"
-       :build `(("make" ,(format "emacs=%s" el-get-emacs) "compile")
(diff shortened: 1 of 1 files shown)


ericdanan:
Since `ivy` does not properly speaking depend on `avy`, an alternative solution that work for me is to leave the dependencies unchanged and only modify the build command by adding the `deps` target:

```elisp
		:build `(("make" ,(format "emacs=%s" el-get-emacs) "deps" "compile")
			 ("makeinfo" "-o" "doc/ivy.info" "doc/ivy.texi"))
		:build/berkeley-unix `(("gmake" ,(format "emacs=%s" el-get-emacs) "deps" "compile")
				       ("gmakeinfo" "-o" "doc/ivy.info" "doc/ivy.texi")))
```

cf abo-abo/swiper/issues/2583
cc @abo-abo

ericdanan:
Well it works but takes a while and adds a folder `~/.elpa` with dependencies, so @lugia-kun's solution is perhaps better.

PR #2757, opened 2020-03-17
Use branch "stable" for c-xrefactory

thoni56:
The "stable" branch will always have passed all tests, so let's use that instead.

PR #2768, opened 2020-07-13
c-xrefactory.rcp: Use stable branch from repo and use new dir for elisp files

thoni56:
The previous PR (https://github.com/dimitri/el-get/pull/2757) never got anywhere, and I wanted to do another small fix so here it is.

- Use "stable" branch from `c-xrefactory` to ensure it builds ok
- The directory for elisp files was renamed to `editors` rather than `env`


thoni56:
Thanks.

PR #2778, opened 2020-07-30
Fix the recipe for deft

vshender:


alexott:
Thank you for submitting it, but it was supersede by #2784

PR #2796, opened 2021-01-10
Add when argument to define-obsolete-function-alias

aaronjensen:
This will be required in Emacs 28


tarao:
The problem should be fixed by #2801 with other Emacs 28 related changes.  Thanks anyway!

PR #2797, opened 2021-01-12
Fix for Emacs 28 or higher versions

syohex:
### Problem

el-get-update, el-get-update-all fails with latest emacs(28.0.50)

### Detail

start-process-shell-command no longer takes rest argument since
Emacs 28. So passing more than three arguments raises
wrong-number-of-arguments exception.

### Reference

Commit
- https://github.com/emacs-mirror/emacs/commit/ecb5ebf156280be1859f181208306e4c55af3e80#diff-e16fa4bdce4bf5bf6122ca90034ac2e3aa60d58085cf20fb1483f4544ad37919

Latest implementation
- https://github.com/emacs-mirror/emacs/blob/ac9c4ca8c9456ea4e0cbfea2317579ac57b13289/lisp/subr.el#L3726-L3737

### DIFF
--- a/el-get-core.el
+++ b/el-get-core.el
@@ -499,9 +499,10 @@ makes it easier to conditionally splice a command into the list.
                         (funcall final-func package)))))
               ;; async case
               (el-get-verbose-message "Running commands asynchronously: %S" commands)
-              (let* ((startf (if shell #'start-process-shell-command #'start-process))
-                     (process-connection-type nil) ; pipe, don't pretend we're a pty
-                     (proc (apply startf cname cbuf program args)))
+              (let* ((process-connection-type nil) ; pipe, don't pretend we're a pty
+                     (proc (if shell
+                               (start-process-shell-command cname cbuf (mapconcat #'identity (cons program args) " "))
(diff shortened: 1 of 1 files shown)


syohex:
`autoload-file-load-name` signature is also changed. However it is difficult to change.

https://github.com/emacs-mirror/emacs/commit/6f36b67e4146ef4610916b7903fd292e1308daf5#diff-92d5169948d378056770f5d1da4c8a70bd4fff0e9d92fe87fa21dd2a8303c13f

bitti:
The third argument of `define-obsolete-function-alias` is mandatory in Emacs 28. Can this line be fixed as well: https://github.com/dimitri/el-get/blob/6138ce5b0a9df779e33ef42eadc4c15a5cf65f39/el-get-status.el#L176?

bitti:
Ah I see that is fixed in https://github.com/dimitri/el-get/pull/2801 or https://github.com/dimitri/el-get/pull/2796. But these are not merged yet. What's the hold-off?

tarao:
The problem should be fixed by #2801 with other Emacs 28 related changes.  Thanks anyway!

Issue #2792, opened 2020-12-28
Replace deprecated cl by cl-lib


stephan-cr:
Since Emacs 27, the package cl is deprecated (and issues a warning: `Package cl is deprecated`). The replacement for cl is cl-lib.

I would like to provide a PR which ports all el-get code to use cl-lib, instead of cl. For this, it would be best to drop support for Emacs versions less than 24.3, because since Emacs 24.3, cl-lib is builtin and doesn't need an extra package from ELPA.

What do you think, can I go ahead?

yyr:
I think its reasonable to drop support for <24.3. its almost 8 years passed since 24.2 is released. I still come across earlier versions of emacs on some linux servers that I use, but again with so many incompatibilities, now a days I tend to install latest version of emacs in home directory on those servers.

So yes, I support such a change.

thoni56:
Emacs now issues a deprecation warning for `cl`, so it is time ;-)

I tried just simple-minded replace all `(require 'cl)` with `(require 'cl-lib)` which seems to be the prefered replacement, but the tests did not run with that change...

stephan-cr:
PR has been merged, therefore, it can be closed.

PR #2802, opened 2021-01-30
Replace cl by cl-lib

stephan-cr:
Since Emacs 27 the package cl is deprecated, the replacement is cl-lib, which is available since Emacs 24.3.

This patch replaces cl by cl-lib and drops support for Emacs versions less than 24.3. Dropping older Emacsen is required, because cl-lib is a builtin starting from version 24.3 and doesn't need an extra package from ELPA.

Testcases for past issues (in test/issues) still contain cl. Most of them seem to be broken and need further investigation.

This patch is tested with `test/run-ert.sh`, which outputs:

```
Ran 10 tests, 10 results as expected, 0 unexpected (2021-01-30 13:24:54+0100, 0.672122 sec)
1 expected failures
```

and manually by daily usage for a month now.

This PR addresses issue #2792.


stephan-cr:
Rebased to current master. Some check for Emacs 24.5 fails with `fatal: origin/master...HEAD: no merge base`.

tarao:
Aha, I guess it is about a workaround replacing `TRAVIS_COMMIT_RANGE`.  I will fix it but it may take some time.

stephan-cr:
@tarao It must be something different, the message `fatal: origin/master...HEAD: no merge base` appears for the successful ones as well.

tarao:
@stephan-cr I think I could fix it now.  It seems that byte-compilation tests should run if there is change outside `recipe/`.

Here are results from a test pull request; I think these are what you have expected:
https://github.com/dimitri/el-get/pull/2838/checks?check_run_id=2813750015

stephan-cr:
@tarao For some strange reason  https://github.com/dimitri/el-get/pull/2838/checks?check_run_id=2813750015 gives:

```
In toplevel form:
methods/el-get-go.el:42:43:Error: reference to free variable ‘el-get-dir’
```

I cannot reproduce it locally, but I fix that nevertheless.

stephan-cr:
@tarao Fixed all compilation errors. The PR doesn't look nice, because it fixes several unrelated errors, due to unreliable CI. I created a follow-up issue to fix the remaining compilation errors.

tarao (approved):
@stephan-cr Great job! Thanks!

PR #2813, opened 2021-02-07
Add ampl mode

vmiheer:


stephan-cr:
I think the correct way to fix upstream, would be:

```elisp
;;;###autoload
(defun ampl-mode () ...)
```

vmiheer:
Thanks, sorry, I missed the location while editing in github website. 🤦‍♂️
Fixed https://github.com/vmiheer/ampl-mode/blob/patch-1/emacs/ampl-mode.el.

stephan-cr:
I wouldn't autoload [`auto-mode-alist`](https://github.com/vmiheer/ampl-mode/blob/patch-1/emacs/ampl-mode.el#L68), see https://www.gnu.org/software/emacs/manual/html_node/elisp/When-to-Autoload.html.

And I would remove

```elisp
(autoload 'ampl-mode "Ampl" "Entering Ampl mode..." t)
```

vmiheer:
@stephan-cr, do we know what's going on with the PRs? This has been open for over a month.

PR #2818, opened 2021-02-10
recipes/lisppaste.rcp seems only available from emacsattic

alexg0:


alexg0:
Ping to please merge this.

PR #2819, opened 2021-02-10
recipes/git-blamed.rcp: git-blame was renamed to git-blamed inside git-emacs

alexg0:


alexg0:
Any concerns with merging this?

PR #2824, opened 2021-02-10
recipe/egg: egg autoloads are too aggressive

alexg0:
egg adds autoloads to find-file-hook, that trigger before egg been initalized.


alexg0:
Ping to please merge this.

PR #2826, opened 2021-02-21
recipes/cmake-mode: Update location of source and website.

ruediger:
Locations have moved.

### DIFF
--- a/recipes/cmake-mode.rcp
+++ b/recipes/cmake-mode.rcp
@@ -1,5 +1,6 @@
 (:name cmake-mode
-       :website "http://www.itk.org/Wiki/CMake/Editors/Emacs"
+       :website "https://gitlab.kitware.com/cmake/community/-/wikis/doc/editors/Emacs"
        :description "Provides syntax highlighting and indentation for CMakeLists.txt and *.cmake source files."
        :type http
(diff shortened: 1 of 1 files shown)


stephan-cr:
@ruediger This is a duplicate of  #2812. But the title is much more descriptive.

vmiheer:
Note the url is slightly different:
```diff
--- https://gitlab.kitware.com/cmake/cmake/-/raw/master/Auxiliary/cmake-mode.el
+++ https://gitlab.kitware.com/cmake/cmake/-/raw/master/Auxiliary/cmake-mode.el?inline=false
```

https://gitlab.com/gitlab-org/gitlab/-/issues/23949

Also who does actual merge? This is the second PR of mine which has competing PR :).

stephan-cr:
> Note the url is slightly different:

Not sure, what's the difference. @ruediger could you clarify?

> This is the second PR of mine which has competing PR :).

I cannot say much about your first PR, but the second one, I would say the recipe name is missing in the title, that's why it could easily overlooked.

ruediger:
Regarding URL: Not sure if the `?inline=false` is necessary. I copied it from the page. But it requires using the `:localname` parameter (otherwise the file will be called `cmake-mode.el?inline=false`). I don't have a strong opinion about that or which PR to merge. I didn't know the other one existed and just wanted to fix the mode.

PR #2838, opened 2021-06-13
(DO NOT MERGE) Test GitHub Actions

tarao:


tarao:
It looks good now.

stephan-cr:
Strangely, running

```shell
$ byte-compile -Werror *.el methods/*.el
```
gives me
```
Source file ‘/home/stephan/.emacs.d/el-get/el-get/el-get-notify.el’ newer than byte-compiled file; using older file

In toplevel form:
el-get-notify.el:24:1:Error: Alias for ‘el-get-growl-notify-path’ should be declared before its referent
```
Neither Travis-CI nor GH do fail for this.

PR #2846, opened 2021-10-08
Update url to clang-format.el since it has moved to GitHub

thoni56:
Since llvm/clang has moved to GitHub that is also where the `clang-format.el` is.

### DIFF
--- a/recipes/clang-format.rcp
+++ b/recipes/clang-format.rcp
@@ -1,7 +1,7 @@
 (:name clang-format
        :description "Clang-format emacs integration for use with C/Objective-C/C++."
        :type http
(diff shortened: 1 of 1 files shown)


alexott (changes requested):
shouldn't it be https://raw.githubusercontent.com/llvm/llvm-project/main/clang/tools/clang-format/clang-format.el ?

thoni56:
From the README at https://github.com/llvm/llvm-project I gathered that llvm has moved to github. Look at the checkout instructions. But I'm confused too... I could not find an official "We're moving!" notice.

alexott:
merged another PR that used the correct URL

PR #2849, opened 2021-11-13
Remove the marmalade archive from the package recipe

stephan-cr:
The marmalade archive is dead for quite some time, that's why it's removed. Furthermore, many regression tests (in test/issues) rely on this recipe and fail, because marmalade is unavailable.


alexott:
thank you!

PR #2854, opened 2021-12-27
Fix Git HTTP smart host detection

stephan-cr:
regression test test/issues/el-get-issue-1920.el revealed that Git HTTP smart host detection is broken. Not all hosts support the HEAD HTTP request method. For example, github.com for HEAD

```elisp
(let ((url-request-method "HEAD"))
  (url-retrieve-synchronously
   "https://github.com/dimitri/el-get.git/info/refs\?service\=git-upload-pack"))
```

responds

```
HTTP/1.1 405 Method Not Allowed
Server: GitHub Babel 2.0
Content-Type: text/plain
Content-Security-Policy: default-src 'none'; sandbox
Content-Length: 0
X-Frame-Options: DENY
X-GitHub-Request-Id: C1DA:12E32:2B6C9D7:302B1C2:61C9A4E8
```

while for GET

```elisp
(let ((url-request-method "GET"))
  (url-retrieve-synchronously
   "https://github.com/dimitri/el-get.git/info/refs\?service\=git-upload-pack"))
```

responds

```
HTTP/1.1 200 OK
Server: GitHub Babel 2.0
Content-Type: application/x-git-upload-pack-advertisement
Content-Security-Policy: default-src 'none'; sandbox
Transfer-Encoding: chunked
expires: Fri, 01 Jan 1980 00:00:00 GMT
pragma: no-cache
Cache-Control: no-cache, max-age=0, must-revalidate
Vary: Accept-Encoding
X-Frame-Options: DENY
X-GitHub-Request-Id: C22C:5E15:3923777:3D822AC:61CA332A
```

Other hosts like git.sr.ht do support HEAD, of course.

Furthermore, the HTTP status code wasn't checked, that's why hosts like github.com would be classified as "dumb" hosts.

This commit checks the HTTP status code, and if the status is not 200 or 304 for the HEAD HTTP request method, it tries GET. HEAD is tried first, because GET might be more expensive for big repositories.

The regression test is adapted as well.

### DIFF
--- a/methods/el-get-git.el
+++ b/methods/el-get-git.el
@@ -15,6 +15,7 @@
 (require 'cl-lib)
 (require 'el-get-core)
 (require 'el-get-recipes)
+(require 'url-http)
 
 (defcustom el-get-git-clone-hook nil
   "Hook run after git clone."
@@ -26,7 +27,7 @@
   :group 'el-get
   :type 'boolean)
 
-(defcustom el-get-git-known-smart-domains '("www.github.com" "www.bitbucket.org" "repo.or.cz" "code.orgmode.org")
+(defcustom el-get-git-known-smart-domains '("www.github.com" "www.bitbucket.org" "repo.or.cz" "git.sr.ht")
   "List of domains which are known to support shallow clone, el-get will not make
 explicit checks for these"
   :group 'el-get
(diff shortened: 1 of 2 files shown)


alexott (approved):
lgtm. thank you

PR #2860, opened 2022-01-30
updated pdf-tools recipe

lpmmo4231:
- fixed wrong load-path and 
- updated minimum required emacs version


vedang:
Please update the minimum version to `26.1`, since `pdf-tools` policy is now to support the latest 3 major versions of Emacs . Thanks!

PR #2863, opened 2022-03-28
Fix peg recipe

takaxp:
The distribution site was changed from emacswiki to elpa.


takaxp:
Note: `org-ql` depends on this pachaige.

alexott (approved):

PR #2864, opened 2022-04-14
Swiper depends on hydra and wgrep

lugia-kun:
According to https://github.com/abo-abo/swiper/blob/7489968257a74f176c0d1de7ec8bd1e2011f0db4/targets/install-deps.el#L17 which is called by `make deps`, swiper depends on avy, hydra, and wgrep.

This change adds the load path of hydra and wgrep for `make compile`. At least I got an error without hydra.


lugia-kun:
#2907 was merged for same issue. Thanks.

PR #2882, opened 2022-11-03
Add / Update recipes for qpdf.el, pdf-tools and elisp-tree-sitter

vedang:
* Add `orgtre/qpdf.el` : A wrapper over `qpdf` for PDF transformations
* Update `pdf-tools` : Clean up the recipe
* Rename `emacs-tree-sitter` to `elisp-tree-sitter` : The repo has been renamed

### DIFF
--- a/recipes/elisp-tree-sitter.rcp
+++ b/recipes/elisp-tree-sitter.rcp
@@ -1,9 +1,8 @@
-;; NOTE: This recipe has been deprecated in favor of elisp-tree-sitter.rcp. Please install that instead.
-(:name emacs-tree-sitter
-       :description "NOTE: This recipe has been deprecated in favor of elisp-tree-sitter. Please install that instead."
+(:name elisp-tree-sitter
+       :description "Tree-sitter bindings for Emacs Lisp."
        :type github
        :minimum-emacs-version "25.1"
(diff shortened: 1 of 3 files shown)


recipes/emacs-tree-sitter.rcp
```diff
@@ -1,10 +0,0 @@
-(:name emacs-tree-sitter
```
alexott, line 1: we need to think how to deprecate it without breaking existing setups... Maybe not remove completely, but add a warning that people need to switch to a new version

alexott (commented):
I'm not sure about removal of `emacs-tree-sitter` without deprecation stage

vedang:
That makes sense, let me push a patch

recipes/emacs-tree-sitter.rcp
vedang: I've un-deleted the recipe and updated the description to say that it is deprecated. Further, I am also updating my fork of `emacs-tree-sitter` to the current latest version of `elisp-tree-sitter`, so people using the fork will continue to get a working version of the repo.

vedang:
I've pushed a patch and replied to the review comment, please review again, thanks!

alexott (approved):
thanks

PR #2888, opened 2022-12-15
Slightly better suggestion of how to set load-path

pedz:
Changed:

`(add-to-list 'load-path "~/.emacs.d/el-get/el-get")`

to

`(add-to-list 'load-path (expand-file-name "el-get/el-get" user-emacs-directory))`

in 9 places allowing the user's emacs directory to be somewhere other than `~/.emacs.d` such as `~/.config/emacs`.


alexott (approved):
lgtm, thank you for pull request

PR #2890, opened 2023-01-03
Ensure ps not to be t

yewton:
Addressing this change: https://github.com/emacs-mirror/emacs/commit/bef1edc9cacb976120dff73b4d7bbdce6ade982b


nevsan:
Ran into this issue as well on emacs-29. I changed the conditional below to ensure we have a list, but this works as well.

PR #2894, opened 2023-02-04
Followup the master -> main transition in the magit recipe

pasja:


pasja:
Fixed by 20a7f652eae29c371f4d3a2573585ad8430516d9

PR #2895, opened 2023-02-06
Emacs 29 fix - make-directory behavior

nevsan:
The behavior of `make-directory` changed such that it will return `t` if the directory exists, rather than always returning `nil`. Here, we update the conditional so that it ensures that `ps` is not a list, instead of just checking if `null`.

### DIFF
--- a/el-get-status.el
+++ b/el-get-status.el
@@ -140,7 +140,7 @@
             (make-directory el-get-dir t)))
          (p-s
           (cond
-           ((null ps) ;; nothing installed, we should install el-get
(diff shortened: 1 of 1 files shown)


renard:
Thanks for this patch. Wouldn't something like 
```elisp
((or (and (< emacs-major-version 29) (null ps)) (not (consp ps)) .... )
```
be more appropriate?

Long story short: keep current behaviour up to emacs below 29 or use your trick otherwise?

alexott:
closing as fixed

PR #2898, opened 2023-03-08
Add new dependency

lispstudent:
No-littering depends also on `compat`. See it source code:

https://github.com/emacscollective/no-littering/blob/main/no-littering.el#L9

```
;; Package-Requires: ((emacs "25.1") (compat "29.1.3.4"))
```


alexott (approved):
thank you!

PR #2899, opened 2023-03-14
all magit/ repos at github now use "main" branch instead of "master"

saffroy:
I recently got an error while installing `magit` with `el-get` and this fixed it.


alexott (commented):
lgtm! thank you

Issue #2881, opened 2022-11-01
Depends deprecated `autoload` package


conao3:
Hi, when I use `el-get` on Emacs-29.0.50 and I get the below message on the Warnings buffer.

```
⛔ Warning (comp): el-get-core.el:25:11: Warning: Package autoload is deprecated
```

I don't know why el-get depends `autoload` but I just wanted to inform you about this.

stephan-cr:
The same happens to the Actions Emacs snapshot builds. All builds currently fail, because byte compilation is done with `(setq byte-compile-error-on-warn t)`.

According to https://mail.gnu.org/r/emacs-devel/2022-09/msg00841.html, the replacement for `autoload` in Emacs 29 is `loaddefs-gen`.

PR #2904, opened 2023-06-16
Upgrade deprecated actions/checkout@v2 to v3

stephan-cr:
Fixes the following Actions warning:

> Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.


stephan-cr:
"snapshot" build fails because of  #2881.

alexott (approved):
lgtm, thank you!

PR #2906, opened 2023-06-18
Remove ERT and "package" installation in CI

stephan-cr:
Since Emacs 23 is dropped for quite some time now, remove installation of ERT and "package", because they are built-in since Emacs 24.


alexott (approved):
lgtm

PR #2907, opened 2023-06-22
Fix swiper build

pasja:
hydra is a build-time dependency of swiper since
40e017dc1bc4655f7c3cf4bbbe3a827ce2fff213 and this patch follows that change in the recipe

Also see https://github.com/abo-abo/swiper/issues/3011


alexott (approved):
lgtm

PR #2911, opened 2023-09-10
Add emacs-openai packages

813gan:


813gan:
marking as draft as installed packages throw warnings during installation.

PR #2955, opened 2024-08-25
Add recipes for ellama and its dependencies: llm and plz

saffroy:
`ellama` provides ways of interacting with various large language models through APIs.

### DIFF
--- a/recipes/plz.rcp
+++ b/recipes/plz.rcp
@@ -1,0 +1,6 @@
+(:name plz
+       :description "An HTTP library that uses curl as a backend"
+       :minimum-emacs-version "27.1"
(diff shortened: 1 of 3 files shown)


recipes/plz.rcp
```diff
@@ -0,0 +1,6 @@
+(:name plz
+       :description "An HTTP library that uses curl as a backend"
+       :minimum-emacs-version "27.1"
+       :type github
+       :pkgname "alphapapa/plz.el"
+       :depends ())
```
alexott, line 6: We don't need `depends` in this case, but it's a minor issue

alexott (approved):

PR #2968, opened 2025-02-18
Add llama as a new magit dependency

pasja:
Since
https://github.com/magit/magit/commit/0a649821007e643ab8c13e770e2a2e10b0389bc4 llama is used in magit, this PR adds it and set it as a magit dependency


alexott (approved):

alexott:
thank you

PR #3012, opened 2026-05-27
Fix el-get-bundle-parse-name

npostavs:
Fixes #3010.

@yewton @alexott I don't suppose either of you remember why you added `(setq s (car spec))` to `el-get-bundle-parse-name`?

(I ended up removing `s` altogether because I think it's just adding noise.)


yewton:
@npostavs I didn't write that part, so it's just a guess, but I suspect `s` was accidentally left in the let binding, and `(setq s (car spec))` was added later as an ad-hoc fix just to silence an unused variable warning.

​I agree it's just adding noise, so removing it completely makes perfect sense to me!

alexott (approved):

PR #3018, opened 2026-06-03
Drop old emacs pre 24.5

npostavs:
Even 24.5 is quite ancient now, but it looks like most of our backwards compatibility code was handling differences in 24.3 package.el vs 24.4, so requiring 24.5 already allows dropping just about all the hacks.


npostavs:
Ah, looks like 28.3 wasn't ever fully released? (https://www.gnu.org/savannah-checkouts/gnu/emacs/history.html does not list it). At any rate https://github.com/purcell/setup-emacs/ doesn't have it.

PR #3022, opened 2026-06-12
Let :repo entries in recipes override package-archive-priorities

### LINKED ISSUE #2832: package-archive-priorities ignored when specifying :repo with type: elpa
My goal is to use melpa stable by default for packages with `:type elpa`, and use melpa (non stable) for select packages.

The way I tried to do this is to set `package-archive-priorities` and specify `:repo` in the receipe of select packages:
```elisp
(require 'package)
(setq package-archives
      '(("gnu" . "https://elpa.gnu.org/packages/")
	("melpa" . "https://melpa.org/packages/")
	("melpa-stable" . "https://stable.melpa.org/packages/"))
      package-archive-priorities
      '(("gnu" . 5)
	("melpa" . 0)
	("melpa-stable" . 10)))
(package-initialize)

(setq el-get-sources
      '(
       (:name magit :type elpa)
       (:name solarized-theme :type elpa
               :repo ("melpa" . "https://melpa.org/packages/"))))

(el-get 'sync)
```

Result: 
- magit is correctly installed with melpa-stable: OK
- solarized-theme is installed with melpa-stable: KO: I would expect it to be installed with melpa archive

Analysis:
https://github.com/dimitri/el-get/blob/d76ac84ae9670de7bf7725b362cafe86688771f9/methods/el-get-elpa.el#L71-L74
`(el-get-elpa-package-id 'solarized-theme)` returns version 1.3.1 from melpa-stable, ignoring the `:repo` I defined in the receipe.

npostavs:
fixes #2832

### DIFF
--- a/el-get-autoloading.el
+++ b/el-get-autoloading.el
@@ -170,7 +170,9 @@ with the named PACKAGE"
           (while (search-forward generate-autoload-section-header nil t)
             (when (member (nth 2 (autoload-read-section-header)) load-names)
               ;; We found a matching section, remove it.
-              (autoload-remove-section (match-beginning 0))))))
+              (autoload-remove-section (match-beginning 0))))
+          ;; Save modifications.
+          (save-buffer)))
       (el-get-update-autoloads package)
       (let ((visiting (find-buffer-visiting el-get-autoload-file)))
         (when visiting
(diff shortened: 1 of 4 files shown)


npostavs:
Hmm, couldn't get get Emacs 24.5 built locally, may have to drop support for this one.

